PR #18, opened 2017-12-14
Update shortcuts on i3 restart

LoganDungeon:
Right now, the shortcut script doesn't run when reloading i3. By changing **exec** to **exec_always** the script runs when reloading i3.

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -35,7 +35,7 @@ exec --no-startup-id sudo mount -a
 #Composite manager:
 #exec --no-startup-id compton --xrender-sync-fence
 #Refresh bash/ranger shortcuts:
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
@LoganDungeon 
This is actually intentional. I only want it to load once in case the files have been edited in the tty or on first start.

If users want the script to reload to new changes, the intention is for them to reload the script by pressing the shortcut Mod+F2.

In fact I might be removing this altogether soon and making reloading fully automatic (as I mentioned in the video, having vim run it on modification of the base files).

LoganDungeon:
@LukeSmithxyz 
Okay, I totally forgot, that this binding exists.

PR #23, opened 2018-02-16
fix Past duration too large

RaitaroH:
May I recommend adding these.

I suppose a better implementation such as getting the refresh rate of the desktop and such. I also drop frames in the encoding so the file isn't huge and has less overhead.


RaitaroH:
Oh jee, that was fast.

PR #28, opened 2018-03-12
fix

RaitaroH:
Thread message queue blocking; consider raising the thread_queue_size option


RaitaroH:
Actually I also get a warning:
```
No pixel format specified, yuv444p for H.264 encoding chosen.
Use -pix_fmt yuv420p for compatibility with outdated media players.
```
But is oke for most people. I don't have any issues,

PR #41, opened 2018-04-24
Document additional urxvt dependencies

### LINKED ISSUE #34: Small warning for using rxvt-unicode instead of st
Since I don't want st as my terminal emulator while this config contains config for urxvt, I tried that instead.
But the URL selection and font resizing doesn't work unless you install `urxvt-perls` and `urxvt-resize-font-git` (AUR), respectively.
This isn't a real issue by itself, but adding a "dependencies required" on the README.md or an option for installing urxvt in LARBS might be a good idea for new users just in case they face the same problem as mine.

I don't know which will work better so its up to you.

CosmoJG:
Resolves #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ There are my dotfiles! The name of the repo, "voidrice", came from the fact they
 
 + i3 (i3-gaps)
 + urxvt (rxvt-unicode) (although these files use [my st build](https://github.com/lukesmithxyz/st) as the default terminal)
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
That isn't really intended to be a dependency list. I might add one, but it's really just supposed to give browsers a list of the dotfiles they can find.

I actually do have a list of deps in the LARBS repo. I might link it in the README here.

PR #55, opened 2018-06-19
Update for multihead systems

marcoacarvalho:
As I use two displays, this script causes an undesirable effect, the padlock image is positioned in the middle of the screens while the green unlock thing is positioned on the center of each screen.

This update takes one separate screenshot per display connected, apply all the effects, merge the padlock image and stich them in order to have one padlock image per screen.

I'm taking the screenshots via imagemagick's import instead of scrot.

As this can be useful to someone else, i'm opening this pull request.

Sorry for my English, It's not my primary language.

### DIFF
--- a/.config/i3/lock.sh
+++ b/.config/i3/lock.sh
@@ -1,13 +1,26 @@
 #!/bin/bash
+#!/bin/bash
 
 # Approximate timeout rate in milliseconds (checked every 5 seconds).
 timeout="10000"
 
-# Take a screenshot:
-scrot /tmp/screen.png
+# Take a screenshot of every screen available:
+xdpyinfo -ext XINERAMA | sed '/^  head #/!d;s///' |
+{
+  while IFS=' :x@,' read i w h x y; do
+    INDEX=$i
+    import -window root -crop ${w}x$h+$x+$y /tmp/head_$i.png
+  done
 
-# Add the lock to the swirled and blurred image:
-[[ -f ~/.config/i3/lock.png ]] && convert /tmp/screen.png -paint 1 -swirl 360  ~/.config/i3/lock.png -gravity center -composite -matte /tmp/screen.png
+# Add the lock to the swirled and blurred images:
+  for i in `seq 0 ${INDEX}`;
+    do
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I'm probably going revert to scrot in my next commits. `import` is painfully slow---it takes about five seconds on my machine compared to the nearly instantaneous `scrot`. Unless you can revise this to a quicker, but still high quality image capture, I'm putting scrot back in.

As for @ofwinterpassed, I'll look into this, although I don't have another screen right now to test it on.

marcoacarvalho:
You're wright about be slower, but isn't the import that is slowing down, is the number of screens detection and unfortunately I can't find a faster way to detect how many screens are attached to the system.

@ofwinterpassed, please run this two commands and provide the outputs:
 xrandr
xdpyinfo -ext XINERAMA | sed '/^  head #/!d;s///'

ofwinterpassed:
@marcoacarvalho I'm at home so right now I can't provide you with the correct info. However I modified the script to the following:
```bash
xdpyinfo -ext XINERAMA | sed '/^  head #/!d;s///' |
{
  while IFS=' :x@,' read i w h x y; do
    INDEX=$i
    xpad="00000${x}"
    xpad=${xpad: -5}
    import -window root -crop ${w}x$h+$x+$y /tmp/head_${xpad}.png
    convert /tmp/head_${xpad}.png -paint 3 -swirl 360 ~/.config/i3/lock.png -gravity center -composite -matte /tmp/head_${xpad}.png;
  done
}
```
Which makes sure the screens are in order.

@LukeSmithxyz Yes it is terribly slow.

ofwinterpassed:
```bash
# Take a screenshot of every screen available:
scrot -m -z /tmp/lock.png
cmd="convert"
xdpyinfo -ext XINERAMA | sed '/^  head #/!d;s///' |
{
  while IFS=' :x@,' read i w h x y; do
		cmd+=" /tmp/lock.png -scale 20% -blur 1.5 -resize 500% -crop ${w}x$h+$x+$y -geometry ${w}x$h+$x+$y ~/.config/i3/lock.png -gravity center -composite -matte"
    done
	cmd+=" /tmp/screen.png"

	eval $cmd
}
```
is a lot faster and I think it will work for multimonitor setups too.

ofwinterpassed:
https://github.com/LukeSmithxyz/voidrice/pull/60

PR #58, opened 2018-06-27
Update .vimrc

ExploXd:
really useful

`set mouse-=a`


ExploXd:
this basically disables visual mode when using a mouse so you can copy.

PR #60, opened 2018-07-01
Faster version of screen lock script by using scrot instead of import

ofwinterpassed:


LukeSmithxyz:
This produces a black screen on my multimonitor setup. Trying to diagnose.

ofwinterpassed:
I updated the script again after learning a bit more about imagemagick and tested it on a dual monitor setup (with two different arrangement of the screens). It seems to work now and it is much faster than using import.

.config/i3/lock.sh
```diff
@@ -4,23 +4,22 @@
...
+scrot -m -z /tmp/lock.png
 xdpyinfo -ext XINERAMA | sed '/^  head #/!d;s///' |
 {
+  cmd="convert /tmp/lock.png"
+	cmdend=""
   while IFS=' :x@,' read i w h x y; do
-    INDEX=$i
-    import -window root -crop ${w}x$h+$x+$y /tmp/head_$i.png
+    cmd+=" -region ${w}x$h+$x+$y -paint 1 -swirl 360"
+    xc=$((x+(w/2)-79))
```
ofwinterpassed: The width of the .config/i3/lock.png is hardcoded to 79 pixels. I guess it would be more elegant to check the width of the image but I didn't have time for this.

LukeSmithxyz:
Well this is what it looks like on my machine. A squirl should really be emanating from the center of each screen or all this blackness appears.

https://lukesmith.xyz/pix/junk/ls1.png
https://lukesmith.xyz/pix/junk/ls2.png

ofwinterpassed:
Is that with the latest version? Mine looks like this (two screens one below the other):

https://www.dropbox.com/s/ce0czjzseztw5b7/screen.png?dl=0

ofwinterpassed:
```
  head #0: 1920x1080 @ 0,0
  head #1: 1920x1080 @ 0,1080
```
that's the relevant part of the output from `xdpyinfo -ext XINERAMA`

This also worked with a slightly different setup (the screens next to each other)
```
  head #0: 1920x1080 @ 0,0
  head #1: 1920x1080 @ 1920,0
```
Though that gave me the picture with the screens next to each other instead. I'd love to make it work on your system but it's hard to fix when I can't reproduce the error.

ofwinterpassed:
Maybe if you add an `echo $cmd` on the line after `eval $cmd` and provide me with your output from `xdpyinfo -ext XINERAMA` I could figure out why you get such a weird result?

LukeSmithxyz:
I'll double check this when I get another chance. I may've been running the original.

CecileGS:
Haven't looked at this i3lock implementation but would it be potentially helpful?

https://github.com/ashinkarov/i3-extras#user-content-i3lock-related

ofwinterpassed:
@CecileGS  It's an interresting script but it doesn't seem to do effects per screen which is the part that's not working on luke's setup.

@LukeSmithxyz No stress. I was just hoping to be able to give something back since I used so many of your tips and learnt so much from your videos.

CecileGS:
I'm guessing this PR has a i3lock that feels responsive enough, and we're just waiting to fix a minor bug? If the swirl transformation is too slow then we can always use a different technique. A frosted pixelation should be good enough for privacy.

ofwinterpassed:
@cecilegs have you tried it. It wouldn't hurt if you did. Testing on more systems is probably a good idea. And yes of course we can modify the effects to whatever you want. I just didn't want to change that since it wasn't the reason for the pr.

CecileGS:
@ofwinterpassed unfortunately I currently don't have the means to  ):  I don't have a dev machine that I can start writing code for PRs let alone test anything here

DuBistKomisch:
@ofwinterpassed trying this out but the swirl seems to happen across the entire image, ignoring any region options. even if I simplify it down to just `convert /tmp/lock.png -region 1920x1080+0+0 -swirl 360 /tmp/screen.png` I get this fully swirled result:

screen

clearly it works in your example though, so wat do? using latest imagemagick 7.0.8 from arch repos

ofwinterpassed:
@DuBistKomisch that seems very weird. have you tried saving the image and running the command manually? My imageMagick version is a bit older than yours (6.9.7-4) and I'm on Ubuntu but I don't think that should matter unless the syntax has changed.

LukeSmithxyz:
At this point, I'm honestly sort of contemplating rewriting the script so that it simply does something computationally simpler on multimonitor systems so we don't have to worry about this. This hasn't really never worked on my multiple displays, and while I'm sure it's possible, I think a simpler solution should be in line of the short term.

DuBistKomisch:
@ofwinterpassed Yes, that was the output of the manually run command. I downgraded to imagemagick 6 and it seems to work. It's mentioned in https://imagemagick.org/script/porting.php that the region option changed in 7. Found [this similar sounding bug](https://github.com/ImageMagick/ImageMagick/issues/1169) about another operation with region which got fixed, so I'll report this as well.

@LukeSmithxyz agree, even on my beefy desktop it takes too long.

P.S. I noticed it should be clearing the region with `+region` or it messes up the padlock icon compositing as well, also only seems to happen in 7 though.

ofwinterpassed:
@LukeSmithxyz Obviously you should make sure that it works in an optimal way. The only reason I did the pr was since my solution was faster than the version using import. I do agree that the swirl takes to much time but I didn't want to change the result of the script. Jsut speed it up. If you just do the paint or some kind of fussy filter instead that could be done to the whole image at once. I don't think positioning the lock is very expensive though if you can get the region issues figured out.

PR #63, opened 2018-07-03
Knowing what music is playing.

ExploXd:
Since I use media keys on my keyboard to operate my mpd I like to know what is currently playing without having ncmpcpp open.

Please ignore that i3 config file. I'm just newbie learning how to git gud at GitHub.

`https://github.com/LukeSmithxyz/voidrice/pull/63/commits/d6b1d3eebb35b5b157e424b73306564297bf7db0`


LukeSmithxyz:
Uh, barring the fact that this pull requests deletes the entire i3 config and the i3blocks config, what exactly is it actually supposed to do?

As it is, the music currently being played by mpd already does show on the status bar...

ExploXd:
Interval in my i3blocks is set to 1 which immediately shows what music is playing at the moment.

LukeSmithxyz:
This is already a feature of these dotfiles. Note that the module takes the signal 11. The `i3mpdupdate` script runs as a daemon on startup in LARBS and it waits for any changes detected via `mpc`. When it detects one (a change in song or pause status, etc.) it signals to i3blocks to update, so all changes in `mpd` are detected immediately and the status bar displays them.

If you installed LARBS and this is not working for you, open a issue.

PR #67, opened 2018-07-06
More appropriate i3 lock icon

bboyle2:
These changes very simply replace the old lock icon with a thinking emoji with a lock composited above his head.
You might say he is wondering what the password is; hmmmm.

The lock.sh script is edited to composite the new lock1.png and apply a -30 deg swirl to give it a nice touch.


LukeSmithxyz:
>thinking emoji
>not a smug anime grill
smh tbh fam

PR #95, opened 2018-08-08
 tpb: start transmission-daemon if it is not running 

TomJansen:
I often forget starting the transmission-daemon when I download something via the 'tpb' script. Now the daemon is started if it was not running already.


LukeSmithxyz:
Good idea. This has happened to me a lot but I never thought to add this feature. Thanks!

PR #96, opened 2018-08-08
Make dmenu use wal colors

ExploXd:
Adding here idea from the issues (more like request).

Also, be sure you add `dmen` and delete `dmenu_run` in the i3 config.

### DIFF
--- a/.scripts/dmen
+++ b/.scripts/dmen
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+# Import the colors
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I don't want to integrate such an *ad hoc* solution. It would also require changing all of the scripts that call dmenu, which would be very inelegant.

As far as I'm concerned, the proper way to fix this is to put out a custom build of dmenu that uses Xresources colors by default. Until then, I think users can put up with it.

PR #97, opened 2018-08-08
Make prompt use wal colors

ExploXd:
Reference to [#96](https://github.com/LukeSmithxyz/voidrice/pull/96).

There are no more actions needed than just merging this.


LukeSmithxyz:
The whole point of the `prompt` command is to have a highly dangerous-looking and bright selection menu. Using the `wal` red usually isn't that. I think it's fine as it is and even if I could use Xresources colors easily, I think this is better.

PR #100, opened 2018-08-09
i3pacman script not creating .pacupgrnum in .larbs directory

termlimit:


LukeSmithxyz:
This was my error. the `~/.larbs/` directory was recently totally removed and I changed the location of that temp file to `~`, but forgot in that one location. I've fixed it now.

PR #115, opened 2018-09-16
Linkhandler always exiting before case statement

kubetz:
In the original code `exit` was executed even when the condition was not met.


LukeSmithxyz:
Silly me.

PR #120, opened 2018-09-28
dunst

bubuntux:
i was looking at dunst and seems like is not necessary to execute manually 

https://wiki.archlinux.org/index.php/Dunst 
'There is no need to start or enable dunst; it is called by systemd when programs send notifications through dbus.'

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -33,8 +33,6 @@ set $hibernate sudo -A systemctl suspend
 # #---Starting External Scripts---# #
 # Setting the background and colorscheme:
 exec_always --no-startup-id wal -i ~/.config/wall.png
-# Starts dunst for notifications:
-exec --no-startup-id dunst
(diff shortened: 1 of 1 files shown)


m3tax:
`notify-send` doesn't work without dunst running, so it might break some scripts.

bubuntux:
hmm i don't run dust and when i notify-send dunst start without having me manually start it

m3tax:
Oh, weird. Never mind that then, I guess my system isn't configured for that.

Rearding the PR, it doesn't really change much because the daemon is very light on resources... and it might also break scripts that check if dunst is running (like [Luke's mailsync.sh](https://github.com/LukeSmithxyz/mutt-wizard/blob/master/etc/mailsync.sh))

bubuntux:
seems like because of those check is not gonna work for luke

PR #139, opened 2018-11-03
replace the copy-to-clipboard hack

vnord:
This is obviously the correct way to do it with xclip as well. I used xsel as xclip doesn't seem to play ball with my clipboard, so if xsel isn't pulled in as a dependency either add it or use xclip instead.


LukeSmithxyz:
Lel totally forgot about this embarrassing line. Thanks for the fix.

PR #142, opened 2018-11-05
Partition labels within dmenumount prompt

dundalkian:
Adds an additional column to parse from lsblk to collect partition label. 

If partition labels exist I probably wrote them and they are probably useful, lets use them I guess.

Using this on system with multiple partitions of similar sizes, nice to be able to type the "human-readable" name of the partition if it exists

This changes dmenu prompt to print:
"[label] /path/ (size)" 
compared to the current:
"/path/ (size)"
(Now only if a label exists, otherwise will print as before)

### DIFF
--- a/.scripts/dmenumount
+++ b/.scripts/dmenumount
@@ -16,7 +16,7 @@ getmount() { \
 	}
 
 mountusb() { \
-	chosen="$(echo "$usbdrives" | dmenu -i -p "Mount which drive?" | awk '{print $1}')"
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
```
awk: cmd. line:1: $2=="part" && $4=="" printf("%s %s (%s)\n",$5,$1,$3)}
awk: cmd. line:1:                      ^ syntax error
awk: cmd. line:1: $2=="part" && $4=="" printf("%s %s (%s)\n",$5,$1,$3)}
awk: cmd. line:1:
No USB drive or Android device detected
```

PR #143, opened 2018-11-07
Prompt shows partition labels

dundalkian:
(Was missing one curly bracket last time)

Adds an additional column to parse from lsblk to collect partition label.

If partition labels exist I probably wrote them and they are probably useful, lets use them I guess.

Using this on system with multiple partitions of similar sizes, nice to be able to type the "human-readable" name of the partition if it exists

This changes dmenu prompt to print:
"label /path/ (size)"
compared to the current:
"/path/ (size)"
(Only if a label exists, otherwise will print as before)


LukeSmithxyz:
Closing this issue since I made big repo changes that will complicate things.gg

LukeSmithxyz:
You can resubmit the PR, but I made some repo changes that erased history complicating this one.

PR #151, opened 2018-11-26
Empty workspace switch

luukvbaal:
Added script which switches to the workspace which a window is being moved to, if that action makes the current workspace empty. Idea from [here](https://github.com/tasinet/i3-switch-if-workspace-empty/blob/master/i3-switch-if-workspace-empty.sh). Modified to not use `jq` for improved performance (~4 times as fast) and no dependencies.

Can imagine this not being the preferred behavior for everyone but for me it greatly improves workflow. Try it out.

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -247,16 +247,16 @@ set $ws9 "9"
 set $ws10 "10"
 
 # switch to workspace
-bindsym $mod+1		workspace $ws1
-bindsym $mod+2		workspace $ws2
-bindsym $mod+3		workspace $ws3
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
Well it's an interesting idea and sure some people could use it, but I definitely do not want this as default.

There are many times I send containers (often running a script) to other workspaces just to get them out of my way to do something else in a new container. This would also cause a lot of frustration to me when using multiple screens.

luukvbaal:
Fair enough, for me it's exactly the other way around hence why I ended up figuring out if it was possible haha.

PR #154, opened 2018-12-06
Handle directories with texclear

Kevin-Mok:
Not sure if you'd be interested in this, but I added some more cases to the `texclear` script. It can be called without any arguments to remove all TeX build files in the current directory or with a specific directory to remove files there. It helps sometimes if I'm just running `pdflatex` in a directory.

*Aside:* On my end, I also created a [list of TeX build files](https://github.com/Kevin-Mok/linux-config/blob/master/txt/tex-build-files.txt) to create the regex from because I couldn't stand that they weren't sorted.  :sweat_smile: But, I put the string literal directly in here.

### DIFF
--- a/.scripts/tools/texclear
+++ b/.scripts/tools/texclear
@@ -3,12 +3,36 @@
 # Clears the build files of a LaTeX/XeLaTeX build.
 # I have vim run this file whenever I exit a .tex file.
 
-case "$1" in
-	*.tex)
-	file=$(readlink -f "$1")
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
This is actually how the `texclear` script originally functioned back in the days. I sort of prefer the way it works now to be honest, you could modify it to be able to write `texclear *tex` if you wanted to though.

I will probably integrate your less ugly alphabetical extension list though.

PR #155, opened 2018-12-13
a better way to turn off/reboot the PC

bigboymate:
Possibly better and easier to read for people new to Linux/GNU using dmenu.

### DIFF
--- a/.scripts/i3cmds/turnoff
+++ b/.scripts/i3cmds/turnoff
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+powerlevel=$(echo "Power off\nReboot" | dmenu -i -p "Power off or reboot?")
+
+if [ "$powerlevel" = "" ]
+then
+	 exit
+fi
+
+if [ $(echo "No\nYes" | dmenu -i -p "Are you sure?") = "Yes" ]
+then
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I actually already have a script that works more like this, but I never pushed it to the repo because I'm still using the `prompt` script in the dotfiles. Here's what the alternative looks like:

```sh
#!/bin/sh
# Gives dmenu menu for shutdown/reboot/hibernate and other system functions.
# You can also feed the choice as an argument manually.

choice="$*"
[ -n "$choice" ] || choice="$(printf "shutdown\\nreboot\\nexit i3\\nhibernate\\nlock screen" | dmenu -p Action?)"

confirm() { [ "$(printf "No\\nYes" | dmenu -nb darkred -sb red -sf white -nf gray -i -p "Really $choice?")" = "Yes" ] ;}

case "$choice" in
	shutdown) confirm && sudo -A shutdown -hP now ;;
	reboot) confirm && sudo -A reboot ;;
	hibernate) confirm && sudo -A systemctl suspend ;;
	"lock screen") confirm && lockscreen ;;
	"exit i3") confirm && i3 exit ;;
esac
```
I think this is a superset of what you have here, so I'll probably just push this script soon instead.

PR #156, opened 2018-12-14
Download video relative to your screen size

rickyjon:


m3tax:
Parsing `xrandr` isn't the best idea (mine doesn't even have "connected" and "primary" on the same lines, and even if it did the resolution isn't the 4th 'word'). Use `xdotool getdisplaygeometry` instead.

LukeSmithxyz:
I think I've decided against merging this. It's an interesting idea, but I often download videos on my smaller laptop screen to watch them later on a big screen. I don't want to make it the default for everyone.

PR #157, opened 2018-12-18
Fix style when showing full weather report

LORDBABUINO:
When clicking on the weather report, the output was print dirty code instead of lines. This commit fixes this error.

### DIFF
--- a/.scripts/statusbar/weather
+++ b/.scripts/statusbar/weather
@@ -3,7 +3,7 @@
 location=""
 
 case $BLOCK_BUTTON in
-    1) $TERMINAL -e less -S ~/.weatherreport ;;
+    1) $TERMINAL -e less -Srf ~/.weatherreport ;;
     3) pgrep -x dunst >/dev/null && notify-send "<b>🌈 Weather module:</b>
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Are you using the most recent version of these dotfiles? This problem shouldn't occur after bfbb806843d770c629d28598c8698a3868e72bd8 I don't believe. At least, it works fine and displays the forecast in color for me. My system isn't a fresh install though.

LORDBABUINO:
I just installed LARBS yesterday, so I think it's updated. The color was working well, but the borders of the table were not.

LukeSmithxyz:
Okay I've checked it on a fresh install and see that it is a problem, but apparently the `-S` option isn't working and the text isn't wrapping. Is this happening on your machine?

LORDBABUINO:
yes

LukeSmithxyz:
Lol okay, this is starting to frustrate me. It works perfectly on my main system just running `less -S` and having those settings in the `~/.profile`: color, no binary warning and no line-wrapping messing it up. But the same settings aren't working on fresh installs. Trying to figure out why.

PR #158, opened 2018-12-19
Duplicate 'e' alias and added music i3block signal

Caedis:
Noticed that the music i3block was not updating and the 'e' alias was not what was set in the .bashrc

### DIFF
--- a/.bmdirs
+++ b/.bmdirs
@@ -8,4 +8,3 @@ vv	~/Videos
 cf	~/.config
 sc	~/.scripts
 mn	/mnt
-e	/etc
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
I've added the first commit, but `lmc` is not intended to update the music module.

There is another script `mpdupdate` I think that should be running in the background that updates the module whenever there is a change in `mpd` whether caused by a cli command or by a song finishing on its own.

If the music module isn't properly updating for you, that means the script has crashed. Is that so? It should start with i3 and tests if it needs to restart whenever i3 is refreshed.

Caedis:
mpdupdate is nowhere to be found.
https://github.com/LukeSmithxyz/voidrice/search?q=mpdupdate

LukeSmithxyz:
Yeah I just saw it deleted in 983e3776586d43c628f02ce99fc1b0d2d04f07f9 where I meant to rename it. It's up now.

PR #161, opened 2018-12-20
Statusbar additions

luukvbaal:
Some statusbar additions, check commit comments.

### DIFF
--- a/.scripts/statusbar/clock
+++ b/.scripts/statusbar/clock
@@ -3,10 +3,9 @@
 date '+%Y %b %d (%a) %I:%M%p'
 
 case $BLOCK_BUTTON in
-	1) pgrep -x dunst >/dev/null && notify-send "$(cal --color=always | sed "s/..7m/<b><span color=\"red\">/;s/..27m/<\/span><\/b>/")" ;;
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
Might be a dumb question, but what is `checkupdates`?

LukeSmithxyz:
Lol yeah I just saw I missed the comment. Okay.

Isn't that just the same as running `pacman -Qu`?

luukvbaal:
The difference is that `checkupdates` doesn't replace the local database used by `pacman` resulting in the fact that it doesn't require root access and is a bit faster. But yeah the output is the same as `pacman -Sy && pacman -Qu`

luukvbaal:
But I suppose, since the statusbar module is showing updates when u would click on it, the repo database is already updated so perhaps using pacman -Qu makes more sense.

LukeSmithxyz:
Yeah my original idea was to have repo syncs done either only manually or via cron with the script I provided. Adding `checkpackages` sort of seems like a redundancy, especially considering that right now, the packages module doesn't even appear if there are no packages to update with the current sync.

So in essence `checkpackages` would only be clickable after the repos were synced meaning that it's be identical to `pacman -Qu` except with an unnecessary wait.

I think for now, if you change it to `pacman -Qu` I'll go ahead and merge it. In the future I might have a permanently packages module in which case `checkpackages` might make more sense though.

luukvbaal:
Yeah sure, done.

PR #164, opened 2018-12-23
Add option to mirror displays with internal display scaling

anders-dc:
This addition allows the user to mirror the desktop across the internal and a single external display. I use this functionality for presentations, where I like to see the slides on my own display so I am turned towards the audience.

The external and internal displays use their preferred resolutions, but the internal display is scaled so that it spans the same desktop area as the external display.

### DIFF
--- a/.scripts/i3cmds/displayselect
+++ b/.scripts/i3cmds/displayselect
@@ -11,7 +11,7 @@ twoscreen() { # If multi-monitor is selected and there are two screens.
     # Mirror displays using native resolution of external display and a scaled
     # version for the internal display
     if [ "$mirror" = "yes" ]; then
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Could you ensure that if the user mirror screens, then attempt to switch to a dual display __without__ mirroring will work properly? As it is, mirroring and then chosing a non-mirroring option for multiple screens does not restore the intended resolutions/placements.

anders-dc:
Thanks for the input, 4dcd3d8 should do the trick

LukeSmithxyz:
Great! Thanks, this is fantastic. This is oddly enough one feature I've never bothered to implement, but I felt that it was sort of needed.

PR #169, opened 2019-01-02
Change `TRUEBROWSER` to `BROWSER` in `linkhandler`

LaisRast:
I searched for `TRUEBROWSER` in the repository but I did not find it in any other file. Thus, I think it should be `BROWSER`.


LukeSmithxyz:
Yeah, it was an older environmental variable now gone. Thanks for noticing.

PR #171, opened 2019-01-03
Simplified ddspawn, no need for sleep

gergely-xyz:
Moved scratchpad show into the i3 config file so dropdowns show right after spawning. This way there is no need to sleep in ddspawn


LukeSmithxyz:
In the future, please test all PR before opening them. This one (1) didn't work and (2) removed the functionality of being able to send additional settings to the terminal emulator from i3.

I've merged it with a minor addition that makes it work with the rest of the system.

gergely-xyz:
My apologies! I did test it as I am currently using it this way. My oversight was that I modified it for gui applications that don't require any additional arguments for my use. I'll get a copy of LARBS on my system and do testing that way in the future.

PR #172, opened 2019-01-04
Warning volume

ExploXd:
Well, sometimes I really enjoy music, and I put my volume usually at 120%. I do like it but my neighbour doesn't, so this is when this comes useful.


LukeSmithxyz:
I don't know if I want to make this universal. On some computers/laptops, you basically always have to have the volume over 100% and it would be sort of an annoyance. My Boomer ears need the high volume.

PR #175, opened 2019-01-07
Add + symbol to the PS1 Prompt in cmd-mode

rickyjon:


LukeSmithxyz:
I don't think this looks particularly elegant. Is it possible to add a symbol _inside_ of the prompt? Maybe colored? It just really sticks out as is.

rickyjon:
As for bash as is, you can't put the symbol inside the prompt as it always display before it. But if you want something more subtle and less intrusive, you can change the text cursor for both modes.

`set vi-ins-mode-string \1\e[6 q\2`
`set vi-cmd-mode-string \1\e[2 q\2`

PR #182, opened 2019-01-14
vifm dir preview, vicmd

luukvbaal:
image
Better looking directory preview (in my opinion) using `tree` pacakge (0.11 MiB). `tree` output might be adjusted to your liking, e.g. different tree depth but this looked fine for most directories on my system.

Also was unable to open/edit files from within vifm which I solved by setting `vicmd=$EDITOR` but perhaps I'm missing a certain commit and you have a different solution for this.

### DIFF
--- a/.config/vifm/vifmrc
+++ b/.config/vifm/vifmrc
@@ -19,6 +19,7 @@ nmap <space> tj
 nmap q ZQ
 fileviewer *.html,*.css,*.py,*.c,*.h,*.sh,*.diff,*.tex highlight -O ansi %c
 fileview */ tree %c -L 1 --dirsfirst
+fileview ../ tree %c -L 1 --dirsfirst
 
 set vicmd=$EDITOR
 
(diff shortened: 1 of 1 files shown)


luukvbaal:
Also http://q2a.vifm.info/269/how-to-add-icons for those interested:
image

luukvbaal:
> Also http://q2a.vifm.info/269/how-to-add-icons for those interested:
I formatted [`ranger_devicons`](https://github.com/alexanderjeurissen/ranger_devicons) source into vifm syntax for completeness. Also made it into a single string since multiple strings greatly reduces vifm startup speed. I first had every icon in a single line as in the `vifm.info` guide and startup took like a full second.
Now I formatted into a single line like [this] and startup seems instant again, unnoticeably slower than without icons and still noticeably faster than ranger.
Could include it in the pull request or open a new one if you're not interested in tree preview Luke.

PR #183, opened 2019-01-14
Android unmount fix

TomJansen:


TomJansen:
Fusermount failed to unmount my phones properly (tested with 2 phones) and umount did work consistently. I came across many people with the same issue with fusermount researching this issue so I thought I should make a pullrequest

PR #189, opened 2019-01-20
Added note to change parameter for urxvt users

LoganDungeon:
It took me a while to realize the error, so this might spare some time of other urxvt users.

### DIFF
--- a/.scripts/i3cmds/ddspawn
+++ b/.scripts/i3cmds/ddspawn
@@ -13,5 +13,6 @@ then
 	i3 "[instance=\"dropdown_$1\"] scratchpad show; [instance=\"dropdown_$1\"] move position center"
 else
 	echo "Window not detected... spawning."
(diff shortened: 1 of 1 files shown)


LoganDungeon:
Maybe you even want to add another if else branch checking which terminal emulator is used.

dawidpotocki:
Well, it can be different with any terminal emulator, so you would have to write down pretty long list. For example, Alacritty uses `--class` flag. Same with GUI applications, it can be different.

LoganDungeon:
I guess you are right.
We can see what Luke thinks about this and he can just delete this PR if he wants.

LukeSmithxyz:
To avoid bloat creep, I make it an issue to not accommodate documentation/case handling of software not in LARBS. This sounds autistic, but otherwise it would be nothing by documentation and hand-holding ad-hockery.

If users are smart enough to change their terminals, they are smart enough to notice this problem and find the solution.

PR #190, opened 2019-01-21
Unplugged displays not being disabled in displayselect

8bitmcu:
This PR fixes scenarios where unplugged displays do not get turned off when changing selections through `displayselect`

In a scenario where `LVDS-1` and `HDMI-1` are active, if the user physically disconnect `HDMI-1` and uses `displayselect`, `HDMI-1` will not get disabled. This PR fixes this issue.


LukeSmithxyz:
Thanks!

PR #191, opened 2019-01-21
Showing the scratchpads initially is not necessary.

LoganDungeon:
The initial show for the scratchpad only makes them show up, when reloading i3, which is a bit annoying.
When calling `ddspawn`, they spawn nontheless.


LukeSmithxyz:
If you remove this line, the windows will not appear when initially spawned.

LukeSmithxyz:
Are you sure about that?

PR #199, opened 2019-01-29
Add font-awesome codepoints to the dmenuunicode menu

malramsay64:
This adds the codepoints for font-awesome to ~/.fontawesome in the same
format as ~/.emojii and modifies the grep search pattern in dmenuunicode
to search both ~/.font-awesome and ~/.emojii including both in the list
of values sent to dmenu.

Removing the ~/.fontawesome file will revert to the previous behaviour.

This requires the fontawesome-fonts package to be installed to get the
additional codepoints.


malramsay64:
I've updated this to now putting the additional file in the .config directory and updated the paths respectively.

PR #205, opened 2019-02-01
Made dunst a user systemd service

mehalter:
Created a user systemd service for dunst. A more concrete setup for something that is constantly running in the background and gives easy access to control and logs from anywhere.

By running `systemd --user enable dunst` it will start on login.

Then you can do `systemd --user status dunst` to check if its running

`systemd --user restart dunst`, `systemd --user start dunst`, `systemd --user stop dunst`, etc.

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -46,8 +46,6 @@ for_window [instance="dropdown_*"] move position center
 # #---Starting External Scripts---# #
 # Setting the background:
 exec --no-startup-id setbg
-# Starts dunst for notifications:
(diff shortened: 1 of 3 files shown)


LukeSmithxyz:
I'd rather not do this for several reasons. While my dotfiles are maintained only for Arch, I try to make them as portable as possible and thus want to avoid using systemd. Here it's not even necessary.

If dunst had a habit of crashing or needed maintaining, I'd contemplate it, but I've never gotten any complaints.

PR #208, opened 2019-02-03
remove a trailing colon from the PATH

iMouath:
The last entry in this path will always have a trailing colon which can cause issues. I've tested it on a new temp user. this should do the trick.

### DIFF
--- a/.profile
+++ b/.profile
@@ -2,7 +2,7 @@
 # Profile file. Runs on login.
 
 # Adds `~/.scripts` and all subdirectories to $PATH
-export PATH="$PATH:$(du "$HOME/.scripts/" | cut -f2 | tr '\n' ':')"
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
How can this cause issues?

iMouath:
> From the bash man pages:
> 
> A zero-length (null) directory name in the value of PATH indicates the current directory. A null directory name may appear as two adjacent colons, or as an initial or trailing colon.
> 
> Putting the current directory in the PATH is generally considered a security risk and a bad idea. It is particularly dangerous when using the root user.

https://stackoverflow.com/a/35740143

You also won't be able to execute some commands for example `find` which is used during package building 

> It's better not to have relative paths in $PATH at all.
> 
> The risk of having . or other relative paths in $PATH is especially heightened when using a utility that automatically changes the directory, which is why find won't let you use -execdir in this situation.

https://askubuntu.com/questions/621132/why-using-the-execdir-action-is-insecure-for-directory-which-is-in-the-path

PR #217, opened 2019-02-07
Respect xdg base directory standards

moham96:
This was made to reduce clutter in the home directory 
see https://www.youtube.com/watch?v=AFtfpluqv14
and  https://0x46.net/thoughts/2019/02/01/dotfile-madness/


LukeSmithxyz:
I don't really understand the point of this. There's nothing that these files have in common that means they should share the same unique directory. It means that every person who uses one of these files will have to add a `~/.config/larbs/` directory.

Having the weather report in `/tmp/` is intentional. It used to be in `~` but I moved it since I *do not* want weather reports to be persistent, so if I reboot my computer tomorrow without internet I won't get a bogus weather report (instead, nothing will appear).

moham96:
> I don't really understand the point of this. There's nothing that these files have in common that means they should share the same unique directory. It means that every person who uses one of these files will have to add a `~/.config/larbs/` directory.

The point is to reduce clutter in the home directory and move them in sub directories.
For me I don't use any kind of file manager, I just use good old `ls` to navigate and since I deal a lot with compilation of software that create dotfiles I gained a habit to navigate hidden files `ls -la`, unfortunately when i navigate the home directory I always get frustrated given how I have OCD about having tons of dot files in the home directory.

what these files have in common is that they are both configuration files used by LARBS in accordance to the [XDG base directory standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

>  $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 

BTW the .scripts directory should also be moved to an XDG directory but I wanted to see if these changes are welcome first.

I don't understand your last point though, the `~/.config/larbs` is created in the repo so the user just clones the repo, what is difficult about adding `~/.config/larbs` ?

> Having the weather report in `/tmp/` is intentional. It used to be in `~` but I moved it since I _do not_ want weather reports to be persistent, so if I reboot my computer tomorrow without internet I won't get a bogus weather report (instead, nothing will appear).

The weather report change was done so the module acts similar to weather modules in other desktop environments and also mobile apps where the weather is cached and if no internet is not available it show the last weather report with  something like "last updated at 05:30 PM"


Of course this is your project and it's up to accommodate my OCD or not, I'm happy to maintain my own fork.

Regards

LukeSmithxyz:
On the `~/.configs/larbs/` point, there is no such program called "larbs" that all of those files have to do with. They have nothing in common. Most people who use this repository do not use the LARBS script (which has nothing to do with any of those files), but they come here from one of my videos looking to get some particular script. That is also why I am extremely reluctant to move files that I have referred to in previous videos. I pay for it in confused emails every time I move something.

That said, I understand the cleanliness issue (don't really care about this "standard" though), but if I do move most of those files it would probably be just to the bare `~/.config/` directory. I've thought it over and I think that that might be a better choice. The bookmark files, for example, are supposed to be general files accessible by any kind of program and it makes sense to put them in a program-neutral directory. Same with the emojis file.

I'm not sure about the alias file though. Most people who have one keep in in `~`. It sort of feels wrong when all the shells that call it will be in the home directory. Maybe I'll move it.

On the scripts, putting a PATH folder in `~/.config/` strikes me as a crime against nature. Do people actually do that? I used to do that when I was a n00b actually. The "proper" location would be `~/(.)bin/` of course, but since I'm a man of the people I use a subtle and comprehensible `~/.scripts/`.

moham96:
> On the scripts, putting a PATH folder in `~/.config/` strikes me as a crime against nature. Do people actually do that? I used to do that when I was a n00b actually. The "proper" location would be `~/(.)bin/` of course, but since I'm a man of the people I use a subtle and comprehensible `~/.scripts/`.

A more 'sane' location for the `.scripts` directory would be in `.local/bin` which is also a standard by systemd and extension to XDG standards, it's also included by default in the `$PATH`  

Regards

LukeSmithxyz:
`~/.local/bin` is not added to PATH by default. In some distros the custom profile adds it, not in Arch though.

dawidpotocki:
> `~/.local/bin` is not added to PATH by default. In some distros the custom profile adds it, not in Arch though.

Are you sure? That was a case on my last Arch install (I don't use Arch anymore, but it wasn't that long time ago), pretty much every systemd distro does that. Also Pythons pip when you use --user flag (or something like that) places files in `~/.local/bin`.

LukeSmithxyz:
Either way, it wouldn't matter if the default Arch profile adds `~/.local/bin` because the profile in this repo which replaces it doesn't. I've tried mking the dir and it still doesn't appear in even PATH on reboot.

LukeSmithxyz:
Okay, I've now moved all of the shortcut files, including `~/.bm*` and the output to other directories
I'm going to close this PR since I'm doing things a little different, but thank you @moham96 for the suggestion.

I moved the readme where you suggested since I suppose it actually is `larbs` specific.

I think I'll move the emoji list as well when I decide where is best or I might possibly integrate it into the script that calls it itself.

I'll be moving some other files and references soon. The notmuch config, gtk2 config, the refer bibliography and others from bare `~`.

PR #219, opened 2019-02-08
update getkeys i3 help

chillinPanda:
Hi @LukeSmithxyz ,

I am starting playing with i3wm and using your LARBS. Big thumbs up and thank you for your work! Since I am new I used `getkeys i3` often.

I noticed that some keybinding were not updated in the `getkeys i3` listing.


LukeSmithxyz:
Great, thanks!

PR #222, opened 2019-02-09
add dmenu to getkeys command

chillinPanda:
Hi @LukeSmithxyz,

I wanted to learn and practice `dmenu` a little bit. So I added `dmenu` to the `getkeys` command. Maybe you like it and merge it.

* one can still launch `getkeys` with a parameter.
* dmenu shows a vertical list of commands for which help files are
available and the user can choose from that list.
* I chose the vertical list so dmenu will attract more attention from the user.

### DIFF
--- a/.scripts/tools/getkeys
+++ b/.scripts/tools/getkeys
@@ -1,4 +1,14 @@
 #!/bin/sh
-cat ~/.config/getkeys/"$1" 2>/dev/null && exit
+# Shows keyboard shortcuts for programs.
+# Program is chosen either by parameter or dmenu.
+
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I'm not really interested in adding this feature. As a note, if you have a terminal command that you want your user to choose an option from, it's probably better to use `fzf` or another dmenu equivalent that works in the terminal itself so (a) you don't need to be in a graphical environment to use dmenu and (b) you don't break the position of focus to the top of the screen.

chillinPanda:
Thank you for suggesting `fzf`. I will have a look into it :)

PR #224, opened 2019-02-12
Background setting from URL feature added!

ceylanb:
With -u argument, the script downloads the file from URL and set the file as background. Script is also working without -u flag like before.


Yash-Garg:
Looks good to me. Nice work!

yochem:
Small thing: maybe change `[ ! -z "$1" ]` to `[ -n "$1" ]`? From `man test`: 

```
     -n string     True if the length of string is nonzero.
```

ceylanb:
@Luke, what was the problem? Why did not you merge it?

PR #229, opened 2019-02-18
Hide the hueg message when running ffmpeg

rickyjon:


LukeSmithxyz:
You need to check your work here.

PR #253, opened 2019-03-23
raw control chars in less, no "ESC" anymore

weidtn:
My weatherreport was full with ESC and other control chars. Using the -R flag for less makes it pretty again.


LukeSmithxyz:
What is your `$LANG`? Does the `-R` maintain color output for you?

weidtn:
`$LANG` should be something German. Can check later.

Yes, the output is still colored with `-R`.

LukeSmithxyz:
Okay. It doesn't seem to break it for English either. Not entirely sure what causes this though.

PR #258, opened 2019-03-30
Prune existing mountpoints from find in dmenumount

TomJansen:
This speeds up the mounting process if you already have drives mounted
containing a large number of folders. Since 2 drives cannot be mounted in
the same folder anyway, these are now excluded from the the find
command, speeding up the mounting process.


LukeSmithxyz:
Thanks, good idea.

PR #264, opened 2019-04-05
Added country indicatior as a status bar option

samerickson:
I realize the implementation is not exactly on par with how you typically make things, but I thought you might be interested in this as an option for VPN users. I personally use this to check what country I am using at a glance.


patosullivan:
This would also require installing `geoiplookup`, it's not installed by default (at least it's not a core utility on Arch, which is what I'm running: https://wiki.archlinux.org/index.php/Core_utilities).

samerickson:
I did not notice that, however the package has an install size of 319.0 KB as per https://www.archlinux.org/packages/extra/x86_64/geoip/
and there are lots of scripts within larbs that are not installed by default i.e transmission. I could add a check if the proper package is installed and display the proper error message in a notification if not.

samerickson:
I did not notice that, however the package has an install size of 319.0 KB as per https://www.archlinux.org/packages/extra/x86_64/geoip/
and there are lots of scripts within larbs that are not installed by default i.e transmission. I could add a check if the proper package is installed and display the proper error message in a notification if not.

samerickson:
Yeah, I found that out a while back as well, specifically when you first boot your system up. My code for the problem is as follows, I haven't had time to push it with university finals going on. 
```
#!/bin/bash

# Gets your public ip address checks which country you are in and
# displays that information in the statusbar

[ "$(cat /sys/class/net/w*/operstate)" = "down" ] && echo "No connection" && exit

addr="$(curl ifconfig.me 2>/dev/null)"
[ -z "$addr" ] && echo "No connection" && exit

grep "$(geoiplookup $addr | sed 's/.*, //')" ~/.local/share/flag-emojis
```

samerickson:
Thanks @davidcastilloc! I deleted the repository that I had created the pull request for as I am no longer interested in the project. My laptop has too many display issues with X11 so I have been in the process for switching to sway, and switch to all wayland compatable tools.

davidcastilloc:
@samerickson It is one last time that X11 is giving problems in computers after so many years of development.
Well! Then Luck with the end of the University, and I wait for your dotfiles for Sway and wayland.
Greetings and a big hug!

LukeSmithxyz:
Okay I merged this, but with some changes, so I guess I have to close this manually. What I changed was:

- The separate flags file isn't actually necessary since there's already an emoji file with all the flags greppable with `flag:`, so I used that as the input and sedded it to be the same to avoid adding a new big file.
- Script is `sh` instead of `bash`
- The module is commented out by default because I assume most people don't use VPNs and know where they live 😉. Like the memory/cpu modules, you can enable them by uncommenting them.
- I renamed the module to `iplocate` since `locate` is an already existing command and that would cause a bunch of weird stuff

I'm sort of curious if it's necessary to `curl` a website to get an IP, maybe things are different for VPNs, I don't use them, but I feel like there should be an easier way to do that...

david-perez:
Any reason to `curl` a third-party website instead of [using DNS](https://unix.stackexchange.com/a/81699) to get the public IP address?

samerickson:
In that case you need to add the dependency `bind-tools` however, I imagine in terms of privacy and security using DNS is better. Using the link you supplied I was able to change the command on my machine to ` dig @resolver1.opendns.com ANY myip.opendns.com +short -4`. Though I am not sure at this point if I should be using IPv6 or IPv4.

PR #266, opened 2019-04-06
Enable Numlock on start

yurisuki:
I like having Numlock having on boot, and I hate having to press Numlock whenever I use the numpad.

### DIFF
--- a/.scripts/config
+++ b/.scripts/config
@@ -1,0 +1,367 @@
+# vim: filetype=i3
+# File originally by Luke Smith <https://lukesmith.xyz>
+
+# This config file will use environmental variables such as $BROWSER and $TERMINAL.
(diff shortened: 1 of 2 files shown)


yurisuki:
Oh, I fucked up, and I've put the i3 config inside .scripts, so if you're going to merge this, take a notice.

LukeSmithxyz:
Well obviously I'm not going to merge this (aside from the error) because not only is it enforcing a particular preference, but every laptop running LARBS would start up typing numbers on the rightward keys.

If you want to do this on your system more cleanly, try this:

```
[ "$(xset q | grep Num | awk '{print $8}')" == "off" ] && xdotool key Num_Lock
```

I'm not quite sure what you're doing by reversing the string twice... Maybe I'm missing something.

PR #272, opened 2019-04-13
Posix shell compatible

Towtow10:


LukeSmithxyz:
This maintains none of the elegance of the original. My preference for POSIX compliance is not so overwhelming. You might as well just manually specify each function to dmenu.

PR #273, opened 2019-04-13
mod+shift+q forcekills window

TomJansen:
In case if a window/program hangs, you can now easily kill it with mod+shift+q


LukeSmithxyz:
Good idea.

Is there any reason why you create a separate script for this?

It appears to work for me to just have:

```
bindsym $mod+Shift+q		[con_id="__focused__" instance="^(?!dropdown_).*$"] exec kill -9 `xdotool getwindowfocus getwindowpid`
```

LukeSmithxyz:
Actually, you should add `--no-startup-id` to that as well.

TomJansen:
True, my script was different first and I did not realize that it could be done inside the config file. I have made a new pull request

PR #281, opened 2019-04-21
Make a bootable drive using this small simple script.

yurisuki:
Ok, I don't know if you find this any useful (I do), and if you want that in your dotfiles, but sometimes this might come handy, and you don't have to look for that command.


LukeSmithxyz:
I don't really see a use for this here. `dd`'s syntax is not so difficult to require a bash wrapper. I don't typically like adding non-essential scripts to this repository anyway.

PR #289, opened 2019-05-01
create 'locker'

elagost:
I just added a pull request to the LARBS repo (https://github.com/LukeSmithxyz/LARBS/pull/126) to add xautolock. It's fairly useful. If you put `xautolock -locker locker` in the i3 config, it'll invoke i3lock after 5 minutes, and if the computer is on battery power, should suspend it 10 minutes after that. If the machine is unlocked (i3lock goes away) or plugged in, it should not invoke sleep.

### DIFF
--- a/.scripts/tools/locker
+++ b/.scripts/tools/locker
@@ -1,0 +1,12 @@
+#!/bin/sh
+# option to use suckless slock as well - just put "slock" for both variables below
+lockprogram="i3lock"
+lockcmd="i3lock -e -f -c 120201"
+$lockcmd && xset dpms force off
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I'm not interested in this functionality.

By the way, be mindful that `[[ ]]` are bashisms, so this script would break on systems with a posix sh since it is marked as `#!/bin/sh`. Use `[ ]`.

Also, `! -z` = `-n`. ;-)

elagost:
Thanks for the reply. Figured since it was useful for me, I'd suggest it, but these are _your_ dotfiles, after all. :+1: 
And thank you for the advice on scripting! Appreciate it.

PR #291, opened 2019-05-04
load ~/.Xresources if available

moham96:
Using `/etc/X11/xinit/xinitrc` as a reference


LukeSmithxyz:
People can add a line for this if they want. As a default, this would cause significant breakage for people who set settings incompatible with `st` before installing. My removing the previous built-in Xresources was intentional.

PR #296, opened 2019-05-09
Fix for spawning scripts that require no args

loumray:
For example, "ddspawn vifm" is generating the failing commad "st vifm -e vifm" without this fix.
This proposed change fixes this issue so that "ddspawn vifm" generates the correct "st -e vifm". It also simplify  the i3 command in my view.
Thanks for sharing this helpful script, I used it everywhere!

### DIFF
--- a/.scripts/i3cmds/ddspawn
+++ b/.scripts/i3cmds/ddspawn
@@ -7,11 +7,13 @@
 
 [ -z "$1" ] && exit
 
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
String indexing is a bashism.

loumray:
I see... Just updated the fix to avoid string indexing.

PR #312, opened 2019-05-24
Dmenu script for monitor selection checks scripts.

stefanocereda:
When using multiple monitors, the current configuration may not be enough (e.g., a vertical screen), which forces to use manual selection and arandr.
Arandr allows to save the configuration in a shell script for future use.
I think it would be handy to be able to select existing scripts from the dmenu monitor selector.

The dmenu now checks for shell scripts under ~/.screenlayouts (which need to be created from arandr) and allows to pick one of them.

### DIFF
--- a/.local/bin/i3cmds/displayselect
+++ b/.local/bin/i3cmds/displayselect
@@ -55,13 +55,13 @@ multimon() { # Multi-monitor handler.
 	esac ;}
 
 loadconf() { # Load an existing configuration file saved from arandr
-	scripts=$(ls ~/.screenlayout/)
+	scripts=$(ls ~/.config/screenlayout/)
 	if [ -z "$scripts" ]; then
 		scripts="None"
 	fi
 	chosen=$(for s in $scripts; do echo $s; done | dmenu -i -p "Select configuration file:") &&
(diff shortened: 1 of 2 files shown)


decantr:
Might be worth having it look for this folder inside `.config` to keep `~` clean.

stefanocereda:
I used ~/.screenlayouts because that's the default folder suggested by ARandR.
I can change that, but then I think it might be better to insert a comment in the document, otherwise it would be difficult to use this feature.

Right now you can use arandr, save a file and get it automatically in the screen selection. If I change the folder you need to know it when saving the configuration.

What you think is better?

decantr:
Perhaps add a note into the [readme.mom](https://github.com/LukeSmithxyz/voidrice/tree/archi3/.local/share/larbs) describing the folder location as in `.config` and not the typical location?

LukeSmithxyz:
Unless there's some other program that is expecting the exact directory `~/.screenlayouts/`, there's no reason for it to have it's own directory in home.

Anyway, this gives me an idea for a more extensible interface though: all original options are separate scripts in a directory and each one is read as a initial option to dmenu. I'll start doing this.

PR #313, opened 2019-05-24
Update .profile

ertwro:
One command less (removes `cut`) when exporting PATH. I know you don't like `find` but is faster here. Is posix, although is not part of Plan9.


LukeSmithxyz:
Let a man live with his irrational hatred of `find`.

PR #315, opened 2019-05-26
mailsync and mpd problems

badmrfrosty:
mailsync does not seem to work. Replacing by the full path .config/mutt/etc/mailsync.sh


LukeSmithxyz:
The shortcut is for the current installable version of mutt-wizard.

PR #316, opened 2019-05-27
Fix 404

rasa:


LukeSmithxyz:
You relinked the archi3 README to the voiddwm programs list which will give the wrong programs.

PR #317, opened 2019-05-27
add dmenu script for power state management under a single command

diegolealco:
You can abscract the lock and exit options into their own functions, scripts, or env variable to be set to whatever you want.

I have my screen lock command in its own script which is then fed to dmenulogout and xxs-lock to keep my lock screen consistent.

There are many ways to do this, I have found this to be very portable whilst uncluttering my i3 and sxhkd bindings.

### DIFF
--- a/.config/sxhkd/sxhkdrc
+++ b/.config/sxhkd/sxhkdrc
@@ -36,9 +36,7 @@ super + Scroll_Lock
 super + Insert
 	showclip
 super + shift + x
-	prompt "Shutdown computer?" "sudo -A shutdown -h now"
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
This has been PRed before, I understand the goal, but I just can't get into it. On most normie computers, there will be different keys/function keys for each command so it makes sense to me.

I'd also like a confirmation check too for safety, as the `prompt` script does, but combining that with a list beforehand sort of gets in the way.

If there ever comes a point that I feel I'm running out of keys, I might make something like this default but I don't feel it's necessary.

PR #326, opened 2019-06-04
Added mpd in .xprofile

Wayloque:
When I launch ncmpcpp, it shows "**ncmpcpp: connection refused**". I had to run `mpd` in terminal manually in order for ncmpcpp to work. Added this to `.xprofile` so it runs automatically every time on boot.


LukeSmithxyz:
Ah oopsie. I had moved this line to the `profile` file, but never committed it to the remote repo. It's there now, thanks for bring this to my attention. I'd rather it start with login rather than the start of the graphical environment with `xprofile` since mpd is theoretically useable in the tty.

PR #332, opened 2019-06-12
Refreshing dwm status bar using signals.

msinkec:
True chads use signals and traps.
No, seriously, it's much more efficient.

### DIFF
--- a/.local/bin/dwmbar
+++ b/.local/bin/dwmbar
@@ -3,7 +3,7 @@
 # This script sets the statusbar with the xsetroot command at the end. Have it
 # started by ~/.xinitrc or ~/.xprofile.
 
-# Handle SIGCONT signals sent by refbar to update the status bar immediately.
+# Handle SIGTRAP signals sent by refbar to update the status bar immediately.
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
Well the true chad solution would have different signals for each module like i3blocks, but that's next on the list.

jmauss:
Hey guys, trying to work on a similar setup. I have "pkill -SIGTRAP dwmbar" in my refbar script, but it doesn't seem to do anything. Once I change it to "pkill -SIGTRAP bash", it seems to update the bar immediately, but I don't want to cause other problems by doing this. How do I get the process to register as dwmbar rather than bash?

dwmbar is located in .local/bin which I have confirmed is in my $PATH. refbar is in the same location, and dwmbar is called in my .xprofile with "dwmbar &"

Thanks!

msinkec:
Can you post your scripts?

jmauss:
They can be found here: https://github.com/jmauss/Scripts

We can either continue discussion on this thread or somewhere on my repo if that makes more sense. Thanks for the help!

PR #334, opened 2019-06-15
Volume buttons support

maxim-h:
Found my keyboard volume buttons not working in larbs (archi3). At least not always, because i feel like they worked in some scenarios on fresh install. Though at that time it was hard to tell, my audio was pretty much inexistent due to the lack of pulse & alsa. Anyway, this got it working. Sorry if it's stupid, i'm very much a newb.

### DIFF
--- a/.config/sxhkd/sxhkdrc
+++ b/.config/sxhkd/sxhkdrc
@@ -80,6 +80,18 @@ super + {_,shift +} bracketright
 super + {_,shift +} bracketleft
 	lmc back {10,120}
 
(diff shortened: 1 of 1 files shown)


decantr:
You can just use the [lmc](https://github.com/LukeSmithxyz/voidrice/blob/archi3/.local/bin/tools/lmc) tool to handle the volume changing for you.

You can also use the `{}` to avoid having multiple entries like so
```
XF86Audio{Raise,Lower}Volume
  lmc {up, down} 15
```

PR #341, opened 2019-06-22
dmenupassword - a script to quickly copy passwords stored in pass to the clipboard

badmrfrosty:
This commit aims to provide a mean to quickly access passwords stored in .password-store directory without using the mouse.
Thanks to this script, you can easily access the password required to login to one of your accounts within a web browser for example : 

- Ctrl+Shift+F1
- Choose your account and press enter
- Ctrl+v in the password field

### DIFF
--- a/.config/sxhkd/sxhkdrc
+++ b/.config/sxhkd/sxhkdrc
@@ -115,5 +115,6 @@ super + F11
 # Network Manager interface
 super + F12
 	$TERMINAL -e sudo -A nmtui
-
+super + Shift + F1
+	dmenupassword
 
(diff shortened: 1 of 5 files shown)


badmrfrosty:
sxhkd -m 1 is necessary to be able to use the sxhkd shortcuts for non QWERTY keyboards (see sxhkd manual)

chambln:
Are you aware of `passmenu`? I believe its executable comes with `pacman -S pass`.

    $ passmenu  # Starts dmenu; you narrow it down to e.g. github
    Copied github.com to clipboard. Will clear in 45 seconds.

badmrfrosty:
Hello,

Thanks for your reply. No I wasn't aware of this little script, that's why I proposed mine. So this can be rejected.

Best Regards,

PR #342, opened 2019-06-25
Added selection of saved display layouts

mehalter:
`arandr` allows you to save display layouts (default save location is `~/.screenlayouts`), so I added those layouts to the main menu of `displayselect` for easy switching between known layouts.


mehalter:
This also could easily be moved to its own menu like a "saved-layouts" menu option.

decantr:
Pretty sure this was already done in #312.

PR #351, opened 2019-07-12
A more elegant way to add ~/.local/bin directories recursively to the path.

chambln:
Just a subtle change to how the directories nested in `~/.local/bin` are added to `$PATH` in `~/.profile`. It now uses `find $HOME/.local/bin -type d` to list recusively only the directories (not files), and GNU `paste` to join the lines together with a colon delimiter.

I've also updated the relevant documentation.

### DIFF
--- a/.local/bin/SCRIPTS.md
+++ b/.local/bin/SCRIPTS.md
@@ -5,7 +5,7 @@ into sub-directories for easy management, and all are seamlessly added to
 `$PATH` with the command below in `~/.profile`:
 
 ```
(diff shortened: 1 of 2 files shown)


ertwro:
You're losing your time, man. He's already rejected similar pull requests. in #239 and #313 
The man's words: https://github.com/LukeSmithxyz/voidrice/pull/313#issuecomment-497974109
Neither ```find``` nor ```paste``` are part of Plan9. Both are Gnu core utils, so they are less POSIX compliant and arguably more bloated. Check it out https://tools.suckless.org/9base/
Here a list of bloated common use software and better alternatives http://harmful.cat-v.org/software/

PR #353, opened 2019-07-29
start i3 only once after the computer is started

moham96:
if the user exited i3 and logged out then logged in, it's up to the user to start i3 again.
This prevent a weird behavior where i want to stop the gui and work in the tty1(when debugging the dot files) but every time i log out and in again the shell starts the gui automatically. 
the /tmp/i3startedonce will be deleted on reboot so the next boot the i3wm will start automatically


decantr:
Is it not simpler to just log into tty2 when you don't want to run X?

PR #355, opened 2019-08-04
Use /usr/bin/env * instead of /bin/*

NilsIrl:
Reasons for:

* This is more portable
* for example without this, these dotfiles cannot work on nixos
* this is the "proper" way

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 stty -ixon # Disable ctrl-s and ctrl-q.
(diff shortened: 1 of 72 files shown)


darthdeus:
@NilsIrl Mostly out of curiosity, but I don't think you need `/usr/bin/env sh`. You definitely do need `/usr/bin/env bash` though. Though it probably doesn't matter, since afaik LSB requires both `/bin/sh` and `/usr/bin/env`. I wonder if there is a system out there that doesn't have `/bin/sh` but has `/usr/bin/env` (even NixOS has `/bin/sh`).

PR #357, opened 2019-08-10
Calcurse i3block & appropriate configuation load

gazny:
New LARBS user fails to load calcurse via i3blocks clock middle click, once loaded the inappropriate configuration would be loaded without any appointments saved from calcurse saved via super + y


decantr:
You've got the cart before the horse here. Just add `-D .config/calcurse` to the `sxhkdrc`.

PR #373, opened 2019-09-13
remove ^t next-unread keybinding

kronikpillow:
the ^t next-unread keybinding is a duplicate keybinding in this config file to the same function with the n keybinding, with bind-key ^t next-unread newsboat shows ^t as the binding when you open newsboat,  but it doesn't work for me for some reason, only n works as the next unread, at least for me, i treid removing the n keybinding completely but it didn't work for me, so I just decided to remove ^t, and now newsboat shows n as the binding

### DIFF
--- a/.config/newsboat/config
+++ b/.config/newsboat/config
@@ -21,7 +21,6 @@ bind-key N prev-unread
 bind-key D pb-download
 bind-key U show-urls
 bind-key x pb-delete
-bind-key ^t next-unread
 
 color listnormal cyan default
 color listfocus black yellow standout bold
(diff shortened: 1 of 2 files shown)


vladdoster:
Why was this closed? Already fixed? If so, could you reference the issue?

kronikpillow:
sorry about that, I'v been messing with the init.vim in my local repository and this is my first time making a git pull request and using local repositories on linux, and i made some mistakes, so i closed the git pull request until i fix things :) but yes, in .config/newsboat/config there is a duplicate keybinding
bind-key n next-unread
bind-key ^t next-unread

newsboat is showint ^t as the keybinding  but it's not working, if i remove the n keybinding it still doesn't work for me, so the only solution i found was to remove the ^t binding, but you guys are better at this then me, so figure it out :)

PR #382, opened 2019-09-29
fixed uberzug, merged variables from voiddwm

kronikpillow:
edited the variable for $FILE to run vu script for image previews to work always... figured since you said your not updating the archi3 branch that much anymore that I should do some work, so I took some parts from the voiddwm branch regarding the dotfiles and merged some of the settings to .profile but I kept archi3 default programs firefox and vifm, which I can change to surf and lf easily if needed, also found that the remaps script or its counterpart settings in .xprofile were technically bloat as the script wasn't used by anything so I updated it to use the same settings from .xprofile and updated .xprofile to use the script instead, and finally, removed the vimwiki keybinding from i3 config and moved it to sxhkd as vimwiki isn't a i3 binding :)


kronikpillow:
found a solution to vifm+uberzug on distrotube and redit and implemented it, you were missing the vifmrun script to launch it with uberzug, vifm now must be run with that script to generate image previews so i edited the variable for $FILE as well to match it, image previews now work .... figgured since you said your not updating the archi3 branch that much anymore that I should do some work, so I took some parts from the voiddwm branch regarding the dotfiles and merged some of the settings to .profile but I kept archi3 default programs firefox and vifm, which I can change to surf and lf easily if needed, also found that the remaps script or its counterpart settings in .xprofile were technically bloat as it wasn't used by anything so I updated it to use the same settings from .xprofile and updated .xprofile to use the script instead, and finally, removed the vimwiki keybinding from i3 config and moved it to sxhkd as vimwiki isn't a i3 binding :)

PR #388, opened 2019-10-14
Remove `exec` from `startx` in `~/.profile`

sarpik:
https://wiki.archlinux.org/index.php/Xinit#Autostart_X_at_login

> If you would like to remain logged in when the X session ends, remove exec.

I found this useful for myself - if the `X` server fails to start,
I'd get thrown out of my session (obviously, because `exec` is used).

This meant that I had to use another user which does not use the `X` server
for troubleshooting. For me, `root` worked, but you might not have access to it,
and this could help.

I don't know if this makes the setup any less secure;
if it does - lemme know.

### DIFF
--- a/.profile
+++ b/.profile
@@ -31,7 +31,7 @@ mpd >/dev/null 2>&1 &
 echo "$0" | grep "bash$" >/dev/null && [ -f ~/.bashrc ] && source "$HOME/.bashrc"
 
 # Start graphical server if i3 not already running.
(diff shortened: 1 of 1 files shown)


ispanos:
Why use another user and not another tty?

sarpik:
@ispanos because since the configuration is messed up, you can log in with as many different ttys to the same user as you want - you'll still get kicked out. Am i missing something here?

reubenkaiser:
Luke's set it up so startx is automatic on login to tty1 only, you can use others and you'll have to use startx to run the graphical environment

PR #395, opened 2019-10-24
ssh script.

eth-man:


.local/bin/i3cmds/sshmenu
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+#$TERMINAL -e ssh `cat ~/.ssh/config | grep ^Host | awk '{ print $2 }' | dmenu`
+
+$TERMINAL -e ssh `cat ~/.ssh/config | grep ^Host | cut -d" " -f2 | dmenu -l 35 -p "SSH to?"`
```
sarpik: Move the grep string into quotes - it does some random stuff without them:

```diff
-grep ^Host
+grep "^Host"
```

Final:

```sh
$TERMINAL -e ssh `cat ~/.ssh/config | grep "^Host" | cut -d" " -f2 | dmenu -l 35 -p "SSH to?"`
```

---

The random output without quotes:

```sh
grep: Documents: Is a directory
grep: Downloads: Is a directory
grep: larbs: Is a directory
grep: Mail: Is a directory
grep: mnt: Is a directory
grep: projects: Is a directory
grep: sync: Is a directory
grep: Videos: Is a directory
```

PR #423, opened 2019-12-03
Implement the best solution to pausing mpv

rickyjon:
I had to use the wayback machine to find the pstree command. No more socat bloat.

### DIFF
--- a/.config/mpv/script-opt/pause_toggle.conf
+++ b/.config/mpv/script-opt/pause_toggle.conf
@@ -1,0 +1,1 @@
+pause_toggle=\
--- a/.config/mpv/scripts/pause_toggle.lua
+++ b/.config/mpv/scripts/pause_toggle.lua
@@ -1,0 +1,22 @@
+-- Adds an additional keybinding to pause an unpaused mpv
+
+local mp = require 'mp'
+local utils = require 'mp.utils'
+local msg = require 'mp.msg'
+
+local opts = {
+
+	pause_toggle = "\\"
+
+}
+
+(require 'mp.options').read_options(opts, "pause_toggle")
+-- main
+-- keybind to toggle pause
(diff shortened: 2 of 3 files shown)


LukeSmithxyz:
Well I was testing this and I have to say I can't get it working and don't have the time right now. `pstree -p | grep "$TERMINAL.*{mpv}"` consistently returns nothing even with multiple mpv instances.

One of the problems is the existences of the `$TERMINAL` variable---I'm not sure why that's needed. I pretty much never open mpv directly from the terminal without `setsid` in a script or something. Even the instances I opened directly from st didn't work pause consistently either. It has something to do with `pstree`'s output getting cut off on the right side. I can't be bothered to figure it out now.

rickyjon:
Alright, I have just realized that the pstree command you mentioned ages ago was specifically for instances of mpv spawned from the terminal that is **audio only**, you can choose to remove the pstree line if you don't care about pausing audio only mpv from the terminal. 
I can't figure out why running the pauseallmpv script outside of a user-interactive terminal doesn't respect export variables, so for now it has a `TERMINAL="st"`, not ideal but it will do much better than having socat installed.

You can also edit the Lua script to be purely hardcoded to backslash key (or has to be modified in the source code for other users like st but without recompiling the entire thing) if you want to save a few extra kilobyte of space and/or amount of files.

LukeSmithxyz:
Maybe you didn't understand my point, requiring any `TERMINAL` variable in there will ensure that _nothing_ or at least mostly nothing will work. Especially using the LARBS dotfiles, you're basically never going to open an instance of mpv that is not independent of the terminal it is spawned from because I always use `setsid` before the command. So in all of my video instances of mpv, there will never, ever be based in an instance of `TERMINAL`.

PR #433, opened 2019-12-14
Add ntfs mounting support

maxim-h:
Additional support for mounting ntfs filesystem drives. Would require adding ntfs-3g to dependencies though I'm not sure where such a list is in LARBS.


LukeSmithxyz:
Isn't ntfs support already implicit in the elsewhere case? I believe I've been mounting a ntfs filesystem with the script as is for a while.

maxim-h:
> 
> 
> Isn't ntfs support already implicit in the elsewhere case? I believe I've been mounting a ntfs filesystem with the script as is for a while.

Yes, you're right. So my problem was, I guess, with the LARBS installer failing to install ntfs-3g on arch. I'll look into that.

PR #446, opened 2020-01-11
Fix Gnome Apps Slow Start Up Issue

goldcoders:


goldcoders:
This happens if you pull up flatpak in your system...
what happens all gnome apps started to fuck up your loading time of all your gnome apps...
im posting it here coz i know someone out there might want to download flatpak 
not everyone is minimalist
but this fix the issue just in case you wanna run flatpak... and face issue of slow start up of all your gnome apps

goldcoders:
its also speed up all gnome apps significantly , the loading times if you added this on your .xinitrc

Basically i dont care if he add this or not, but im making a pull request just in case someone face up this issue with gnome apps...

vladdoster:
Gotcha, in that case it might be worthwhile opening an issue and reiterating what you've stated in an issue. But am curious to see what luke does with this PR.

goldcoders:
Sooner or later you will download a gnome apps,
void rice uses some gnome apps also...
like gnome-keyring...

This wouldnt hurt someone if they place this on .xinitrc

But just in case you need a reference here 

References :
https://bbs.archlinux.org/viewtopic.php?id=230036
https://github.com/flatpak/xdg-desktop-portal-gtk/issues/72

goldcoders:
i wouldnt consider 1 line of code a bloat lol hahaha

vladdoster:
Well it doesn't seem that flatpak in particular is used in voidrice. The aur and pkgbuilds seems to do what flatpak does.

 Does this effect non-gui apps, I think that is what gnome-keyring is.

vladdoster:
Also when we say delay, are we talking about seconds or minutes?

goldcoders:
mostly apps delay are 30 secs but ,i have experience beyond 1 minute 

yeah GUI apps have perf issue

here is an example:

https://youtu.be/jEkb27cZv6U

kronikpillow:
I wonder what the gnome-keyring is used for in LARBS in the first place, does luke use it for SSH and GPG keys?

kronikpillow:
the dbus-update-activation-environment --systemd DBUS_SESSION_BUS_ADDRESS DISPLAY XAUTHORITY is not only needed for flatpak, it's needed for some apps like Skype and so on, so that they would actually display the gnome-keyring pass phrase, but I dunno if it is needed for the LARBS build, as Luke doesn't seem to be using them, i mean these kind of minor tweaks can be done by a user on a per user basis/need

PR #448, opened 2020-01-12
fix(i3 config)

vladdoster:
Removed redundant keyboard shortcut that can be found in the sxhkdrc config


vladdoster:
see [here](https://github.com/LukeSmithxyz/voidrice/blob/master/.config/sxhkd/sxhkdrc#L53)

PR #451, opened 2020-01-14
corrected file extension error with Rmd produced pdf's opening with opout

tallguyjenks:
casing is sensitive on the rmd/Rmd files, without the uppercase option opout was not functioning for me on default LARBS with thinkpad x220. 

tested on home system and verified that functionality as intended is achieved with the `*.Rmd` option is added to the case statement


tallguyjenks:
resolved with #454 and additional changes made for that issue.

PR #456, opened 2020-01-20
Further ~/ cleanup

Daja177:
This pull request will move more files and folders from the home directory to locations more XDG compliant.

When downloading files using Wget it saves known hosts in the .wget-hsts file. There isn't an environment variable to change its location, but it is possible to do that in Wget's config file. Unlike the hsts file, this file's location can be changed with an environment variable which is why it was added here.

Moving the .Xauthority file may cause issues with LightDM, but it should be fine since LARBS doesn't use a display manager by default. Feel free to remove this part if this is a concern.


LukeSmithxyz:
My autism is yet further soothed.

vladdoster:
This commit is so good. ~/ is looking spick and span. Great job on this

LukeSmithxyz:
I reverted the gnupg variable. That had caused some trouble before actually when I had tried it, so I think it'll be safer to keep it in home for others.

LukeSmithxyz:
Actually, the alsa move has caused some problems; `amixer` and `alsamixer` do not work with it due to an error I can't figure out. Of course, you can specify in `/usr/share/alsa/alsa.conf` manually where you want it to look for configs, which I might just make the LARBS scripts do.

Daja177:
Hmm that's interesting. I've been using the GnuPG variable for a few months now and so far it hasn't caused me any issues.

Not sure why amixer and alsamixer aren't working correctly. The environment variable was able to redirect the config's path according to my tests (although I never did any audio tests since I'm not currently using alsa), so it shouldn't be causing any problems. Maybe I'll report this issue to them later.

Thanks for the feedback! @LukeSmithxyz

PR #457, opened 2020-01-21
fix(remove redundant nmtui bindings)

vladdoster:
The nmtui binding had 2 binding in the i3 config even though it can be found in [sxhkd config](https://github.com/LukeSmithxyz/voidrice/blob/master/.config/sxhkd/sxhkdrc#L38)


LukeSmithxyz:
I'm removing sxhkd from use with i3.

PR #458, opened 2020-01-21
fix(hibernate in sxhkd)

vladdoster:
I'm not too sure what `zzz` is, but in i3 it is achieved a separate way. Hibernate should work now.

### DIFF
--- a/.config/sxhkd/sxhkdrc
+++ b/.config/sxhkd/sxhkdrc
@@ -125,7 +125,7 @@ XF86PowerOff
 XF86Calculator
 	$TERMINAL -e bc
 XF86Sleep
-	prompt 'Hibernate computer?' 'sudo -A zzz'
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
zzz is for runit devices. What I'd really like is a hibernate bind universal for init systems. Probably some obvious command I've overlooked.

drake-newell:
> zzz is for runit devices. What I'd really like is a hibernate bind universal for init systems. Probably some obvious command I've overlooked.

As far as I know, there is. It's called pm-utils (it's in the aur).

LukeSmithxyz:
Things you have to install from the AUR by definition aren't universal.

drake-newell:
> Things you have to install from the AUR by definition aren't universal.

Forgot to mention that it's also in the void main repos and debian main repos.  It's a fair bit more complicated that zzz, but it works with any init system.

PR #463, opened 2020-01-26
Add vaapi recording functions

mhdzli:
Hardware video acceleration makes it possible for the video card to decode/encode video, thus offloading the CPU and saving power. 
There is a brief reference with practical useful information in [Arch wiki](https://wiki.archlinux.org/index.php/Vaapi).
These changes mostly inspired by [Kai Henry](https://github.com/kaihendry/recordmydesktop2.0) and [Simon Hugh's](https://github.com/simonhughcom/dmenu-bin/blob/master/recordmenu) scripts.

### DIFF
--- a/.local/bin/dmenurecord
+++ b/.local/bin/dmenurecord
@@ -118,7 +118,7 @@ audio() { \
 	}
 
 askrecording() { \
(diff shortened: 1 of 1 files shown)


mhdzli:
Comparison of CPU usage in video encoding with GPU (using #vaapi) and CPU:

https://lbry.tv/@mzeinali:c/vaapi:8

Issue #465, opened 2020-01-27
zsh no history


eth-man:
please add zsh history to zsh config, thank you

vladdoster:
You add it? @eth-man

vladdoster:
@eth-man is a brainlet, so I had to step up and be a chad XD

PR #466

kronikpillow:
@eth-man ?? why don't you add it your self? It's very simple

edit your .zshrc and add
HISTSIZE=10000
SAVEHIST=10000
HISTFILE=~/.config/zsh/.zhistory

or something .... geeez

vladdoster:
> @eth-man ?? why don't you add it your self? It's very simple
> 
> edit your .zshrc and add
> HISTSIZE=10000
> SAVEHIST=10000
> HISTFILE=~/.config/zsh/.zhistory
> 
> or something .... geeez

I already added it and opened a PR. Don't offend the brainlet XD

Daja177:
@kronikpillow Sure it's easy to add it by yourself, but at least he reported this error. The absence of those lines was likely not intended by Luke, so it's not necessarily a bad issue. Although it would've been more helpful to create a pull request instead, it did result in the creation of @vladdoster's pull request.

eth-man:
im not even using zsh ,you brilliants :)

PR #466, opened 2020-01-27
fix(added zsh history)

vladdoster:
ref #465

Added zsh history


Daja177:
I'm guessing you accidentally pull requested the wrong branch; 208 commits is a bit too much for one single pull request. Most of these don't even relate to each other. In the case that it wasn't a mistake, I'd suggest separating these into different pull requests.

Besides, many of these commits like 52f62a56b54880085813625b3b96ab97c33f281c and 730205258e13ee9f4a7dd47e86218e64d2ae10e6 removes config files from programs currently in LARBS, and in commits like ba4e205e5427edf814fb910e699dd0d030f5e75f you're adding config files from programs Luke may not even have touched before.

Other than that, aa2ff4f2adc7182e7a58f58f7df57536cc94868b is a good commit. I'm not sure why Luke's `zsh` config didn't have that before, especially since Luke had that in his video about `zsh` a while back. I couldn't find a commit that removed those lines, so it's likely just a mistake on Luke's part.

vladdoster:
You right, whoops.

PR #460, opened 2020-01-21
Remove lmc references in i3 config

Daja177:
The `lmc` script was removed in 2e2200b4b96436fbfee59c74bfeb9f7e417a84df, but there still exist references to it in the i3 config. This pull request resolves #459.

I have yet to test for bugs since I don't use ALSA.

PR #471, opened 2020-01-30
Fix i3 audio keybindings

### LINKED ISSUE #470: i3 sound commands are not working
I was having troubles using all the sound commands on i3, I changed the ";" with "&&"  in the i3 configuration file and that solve the issue (by example: bindsym $mod+Shift+m		exec --no-startup-id amixer sset Master toggle **&&** pkill -RTMIN+10 i3blocks). It is necesary make that change or I'm missing a dependency or additional configuration?

Daja177:
In #460 I attempted to replace the `lmc` references with `alsa` and `pkill` commands, but I used the wrong syntax when combining these two. Because of this, none of those keybindings work currently. My bad.

This pull request fixes #470.

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -55,7 +55,7 @@ bindsym $mod+grave		exec --no-startup-id dmenuunicode
 ##bindsym $mod+asciitilde
 
 #STOP/HIDE EVERYTHING:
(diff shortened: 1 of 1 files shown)


Daja177:
It'd be great if you could take a look at this whenever possible @LukeSmithxyz

PR #472, opened 2020-01-31
KBLayoutManager now aids multi-layout keyboards.

MYDavoodeh:
It has numerous problems for the moment (such as slow switching and no variation support) but I think it's a necessary feature for a desktop to have.

`$KBLAYOUTS` holds the list of languages to cycle through.
The value of it must be space separated.

As an example I'm using Farsi/Persian (ir).

### DIFF
--- a/.local/bin/kblayoutmanager
+++ b/.local/bin/kblayoutmanager
@@ -1,0 +1,32 @@
+#!/usr/bin/env sh
+# An adaptation of https://github.com/porras/i3-keyboard-layout/blob/master/i3-keyboard-layout by MYDavoodeh
+set -e
+
+get_kbdlayout() { setxkbmap -query | grep -oP 'layout:\s*\K(\w+)' ;}
+
+set_kbdlayout() {
+	# FIXME what to do for variants?
+	setxkbmap "$1"
+	pgrep i3status | xargs --no-run-if-empty kill -s USR1 # to tell i3status to update
+}
+
+cycle() { #TODO: Optimize cycling time
+	current_layout=$(get_kbdlayout)
+	eval last="\$$#"
+	# If current_layout is the last one OR current_layout is not in the given layouts, set the first one in cycle as the initiator. Else find the next.
+	{ [ "$last" = "$current_layout" ] || ! echo "$*" | grep -qo "$current_layout" ;} && set_kbdlayout "$1" || set_kbdlayout "$(echo "$*" | cut -d' ' -f "$#")"
+}
+
+
+subcommand="$1"
+helpmsg="Please specify one of: get, set <layout> or cycle <layout1> <layout2> ... <layoutN>"
+
+shift || (echo "$helpmsg" && exit)
+case $subcommand in
+	"get") echo get_kbdlayout ;;
+	"set") set_kbdlayout "$1" ;;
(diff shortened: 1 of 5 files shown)


.local/bin/kblayoutmanager
```diff
@@ -0,0 +1,32 @@
...
+	# FIXME what to do for variants?
+	setxkbmap "$1"
+	pgrep i3status | xargs --no-run-if-empty kill -s USR1 # to tell i3status to update
+}
+
+cycle() { #TODO: Optimize cycling time
+	current_layout=$(get_kbdlayout)
+	eval last="\$$#"
+	# If current_layout is the last one OR current_layout is not in the given layouts, set the first one in cycle as the initiator. Else find the next.
+	{ [ "$last" = "$current_layout" ] || ! echo "$*" | grep -qo "$current_layout" ;} && set_kbdlayout "$1" || set_kbdlayout "$(echo "$*" | cut -d' ' -f "$#")"
```
KEZ-, line 17: Nice addition, but it only works with two or fewer layouts defined.
`set_kbdlayout "$(echo "$*" | cut -d' ' -f "$#")"` always selects the last element in list.
It could be fixed with something like this:
`set_kbdlayout $(echo "$*" | sed -E "s/^.*$current_layout ([a-zA-Z]*) .*$/\1/")`

MYDavoodeh: Thanks for your comment.
Indeed there is a problem... It's funny I haven't noticed it in years...

I'm afraid I couldn't make your suggested solution to work... (I'm just a novice sorry)

But I'll work on it tomorrow in order to fix it...
Sorry for bad English

.local/bin/statusbar/lang
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+# A simple module to print current KBLAYOUT
+# Works best with my script of KBLayoutManager.
+# This script sends the update signal and it is bound in i3/config.
+lang=$(setxkbmap -print | grep xkb_symbols | awk '{print $4}' | awk -F"+" '{print $2}')
+[ "$lang" = "us" ] && color="#00ff00" || color="#ff0000" # This line is added to add awareness and ease to detect the us layout in a glance to be ascertain that your shortcuts and keystrokes will remain responsive # FIXME don't be coloured for us variations
```
MYDavoodeh, line 6: English! :facepalm:

.local/bin/kblayoutmanager
KEZ-: This works for me, I had an extra space in my original comment :smiley: 
```suggestion
	{ [ "$last" = "$current_layout" ] || ! echo "$*" | grep -qo "$current_layout" ;} && set_kbdlayout "$1" || set_kbdlayout $(echo "$*" | sed -E "s/^.*$current_layout ([a-zA-Z]*).*$/\1/")
```

MYDavoodeh: Thanks a lot...
I've tested it, and it's pretty robust!

I'll edit my files and post them in another issue... Seems like I've deleted my fork a long time ago...

MYDavoodeh:
This issue is being closed in favor of #634

PR #475, opened 2020-02-03
Update dmenuunicode

LinusCDE:
Using the script `dmenuunicode` from terminal and with a typical graphical environment reveals some things that can be made better.

### DIFF
--- a/.local/bin/dmenuunicode
+++ b/.local/bin/dmenuunicode
@@ -11,8 +11,8 @@ chosen=$(grep -v "#" ~/.local/share/larbs/emoji | dmenu -i -l 20 -fn Monospace-1
 
 c=$(echo "$chosen" | sed "s/ .*//")
 echo "$c" | tr -d '\n' | xclip -selection clipboard
-notify-send "'$c' copied to clipboard." &
+notify-send -a `basename "$0"` "'$c' copied to clipboard." &
 
 s=$(echo "$chosen" | sed "s/.*; //" | awk '{print $1}')
 echo "$s" | tr -d '\n' | xclip
-notify-send "'$s' copied to primary." &
+notify-send -a `basename "$0"` "'$s' copied to primary." &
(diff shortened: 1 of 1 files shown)


Daja177:
@LinusCDE Regarding the first commit 3edf145fe82b910dd05eae5a182f135296ab3556, the `xclip -h` line's purpose is to make sure `xclip` is installed. It disposes of the standard output since it doesn't need it to continue the program. The reason why it doesn't also dispose of the error output is because it's useful for telling why the script exits. If the script doesn't run correctly, it's much easier to debug it by running it in the terminal if it outputs the error there.

The second commit 090e5b1d6eae3d6c5cdb279f8e4b6e660c245998 however looks great

LinusCDE:
Can you even configure xclip when wanting to get the help page?
In the case of installation I would recommend using
`if ! which xclip 2>/dev/null >&2; then exit; fi` or `which xclip 2>/dev/null >&2 || exit`. That would do the job more specifically. If you wanted to know why a script failed I would suggest to

- Exit with another exit code than 0 (default). e.g. exit 1 to indicate that the script failed
- Add a message indicating the reason of it failing.

You could do this either as
`which xclip 2>/dev/null >&2 || (echo 'ERROR: Missing xclip'; exit 1)`
or using a method to streamline failing and shortening multiple lines with that approach
```sh
function error { echo "ERROR: $1"; exit 1; }

which xclip 2>/dev/null >&2 || error 'Missing xclip'
```

That would probably a lot cleaner.

But if you want I can also just revert that commit in question.

LukeSmithxyz:
If you want to add an exit code of 1, that's fine and make sense. I don't know why you want to add some custom error message and function while suppressing the error message that the shell would naturally give when running `xclip` if not installed.

LinusCDE:
Probably because I don't agree that you should run the program to test its output. I treat the script as a program. And as such it should follow the unix philosphy "No news is good news". In my opinion the script should follow all or most guidelines to the outside (stdin/stdout/stderr). How it's done on the inside (source code) is secondary.

Having the script give a specific and useful error to that script (that it is missing that program) is as good as getting the shell-specifc error that the command could not be found. Having one option print useless stuff in the process is the deciding factor for me.

Never mind, it's your repo in the end. If you wish to use that script in your way, I'll revert that commit. No way rambling about how a script should behave.

PR #477, opened 2020-02-04
Use nvim for coloured man pages.

iamantu93:
You can even quit using q instead of  :q.

### DIFF
--- a/.profile
+++ b/.profile
@@ -29,14 +29,7 @@ export PASSWORD_STORE_DIR="$HOME/.local/share/password-store"
 export DICS="/usr/share/stardict/dic/"
 export SUDO_ASKPASS="$HOME/.local/bin/dmenupass"
 export FZF_DEFAULT_OPTS="--layout=reverse --height 40%"
-export LESS=-R
-export LESS_TERMCAP_mb="$(printf '%b' '[1;31m')"
-export LESS_TERMCAP_md="$(printf '%b' '[1;36m')"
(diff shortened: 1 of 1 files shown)


Daja177:
It's an interesting idea, but I'm not really sure I necessarily see the benefit of this.

Since man-pages are meant to be read and not necessarily written to, I find that reading them using `vim` is more of a hassle than using `less` due to the difference in how you scroll in those programs. I've also noticed that the loading time when opening man-pages is longer when using your suggested environment variable. Besides, the environment variables in Luke's `.profile` file already gives `man` colors, although admittedly it's not as detailed as `vim`'s syntax highlighting.

It'd be great if you could provide some more information in regards to the benefits of using this instead, in case there's something I've missed.

iamantu93:
Luke and everyone like the vim keys so much. If you use neovim as a manpager you can use vim keys to navigate. Instead of several environment variables you can achieve colour manpages using only one variable. I haven't noticed any longer loading time.

Daja177:
@iamantu93 `less` uses vi-keys though (at least some of the navigation ones). You scroll up and down with `k` and `j`, half a page up and down with `u` and `d`.

LukeSmithxyz:
This doesn't really yield any advantage. And it certainly is laggy on my system. Additionally, less is a program used for many different purposes so if I did abandon it for reading manuals, I wouldn't want to remove those variables for coloring it in other uses.

PR #480, opened 2020-02-06
Fixed failed compiling of .mom documents

gmoneymagna:
1. Groff and pdfmom commands to create pdf files were being halted by a trivial error, a known bug when converting .mom to .pdf so I had the script ignore this error.
2. I replaced the groff command with the more simple `pdfmom`

### DIFF
--- a/.local/bin/compiler
+++ b/.local/bin/compiler
@@ -24,7 +24,7 @@ textype() { \
 
 case "$file" in
 	*\.ms) refer -PS -e "$file" | groff -me -ms -kept -T pdf > "$base".pdf ;;
-	*\.mom) refer -PS -e "$file" | groff -mom -kept -T pdf > "$base".pdf ;;
+	*\.mom) refer -PS -e "$file" | pdfmom > "$base".pdf 2>/dev/null ;;
 	*\.[0-9]) refer -PS -e "$file" | groff -mandoc -T pdf > "$base".pdf ;;
 	*\.[rR]md) Rscript -e "require(rmarkdown); rmarkdown::render('$file', quiet=TRUE)" ;;
 	*\.tex) textype "$file" ;;


Daja177:
This looks great, but I'd suggest that you add the `-kept` part from the previous `groff`command to the new one, since `pdfmom` accepts all the same options as `groff`. Those options are used to precompile the groff document with other groff programs such as `preconv` and `eqn`, which are definitely useful when working with larger documents.

gmoneymagna:
> This looks great, but I'd suggest that you add the `-kept` part from the previous `groff`command to the new one, since `pdfmom` accepts all the same options as `groff`. Those options are used to precompile the groff document with other groff programs such as `preconv` and `eqn`, which are definitely useful when working with larger documents.

I added it back to my own script, thanks for the tip. I'm trying to figure if I can edit a commit or if I have to submit a new one.

Daja177:
@gmoneymagna It is possible to edit that commit if you're working with git locally by using git's soft reset feature. Soft reset the commit, edit the file, commit it again and then force push it to the remote repository. If you're already working with git locally this will be quite easy to do, but I'm unsure of how to do this from the browser.

However it might be easier to just make a new commit to your branch with these changes, especially if you're only working with GitHub on the browser. Besides, force pushing shouldn't be used lightly since it can cause problems when other people are working on your branch. It shouldn't be an issue in your case since you're the only one working with your branch, but it might be something to consider when working on larger projects.

Also while this isn't really related to the actual content of the pull request—nor is it of great importance, it'd be great if you used an active voice in future commit messages and pull requests. For example, "add" instead of "added" and "merge" instead of "merged". The reason for this is because you want to think of your commit as expressing intent to introduce a change, and it contributes to consistency since most already do this.

PR #492, opened 2020-02-29
Add random wallpaper function to setbg 

mhdzli:
Set a random wallpaper from a directory


LukeSmithxyz:
I've added a commit db46b82f7f0fe7d8bb76246e1e4b9947f4187012 inspired by this that does it a little more elegantly. Give `setbg` any directory name and it randomly chooses a wallpaper.

Be careful though, your additions add bashism without changing the shebang to bash, so this would crash on systems with a posix shell. You can always use `shuf` to get a random file though.

PR #499, opened 2020-03-16
Just find pictures

mhdzli:
Filter the pictures in directory by file extensions


mhdzli:
Wow that was fast.
Thanks for your guidance by the way. :+1:

PR #502, opened 2020-03-22
Update vimrc

nerrufam:
Cleans up listing of archives, adds new archives, and ueberzug can display the first frame of a gif.


kronikpillow:
ur updating  vifmrc not vimrc mate

PR #515, opened 2020-03-25
Add option for an additional battery block

tallguyjenks:
if users have a thinkpad with a slice battery this will let them see both batteries using the same block script in their status bar


tallguyjenks:
resolved with 39bde79

PR #516, opened 2020-03-28
tool for opening scripts shorter and etc

narukeh:
No need to find your script and then open them. This opens your script automatically, by finding its absolute path. It is not limited to local scripts, just make sure it is not a binary.
If it is a binary exec, it can find out which package owns it, and If it isnt a exec, (lib or share) give its full path. Examples:
$ dcf p i3            # this will tell you that "i3-gaps" owns the file "/usr/bin/i3"
$ dcf pl /usr/lib/pkgconfig/zathura.pc    # tells you that its owned by zathura
$ dcf v setbg    # this will open ~/.local/bin/setbg in nvim
$ dcf c -n lmc     # prints out ~/.local/bin/lmc with numbered lines
$ dcf h -15 shortcuts    # prints the first 15 lines of ~/.local/bin/shortcuts
$ dcf t -1 setbg   # prints the last line of ~/.local/bin/setbg
$ dcf l -S volume  # less -S ~/.local/bin/statusbar/volume
$ dcf g -i home weather   # case insensitively grep home from the script weather
$ dcf weather     # works as 'which', only tells you the location of that command

The name stands for "Do command file", but if you have a better one, change it.

### DIFF
--- a/.local/bin/dcf
+++ b/.local/bin/dcf
@@ -1,0 +1,31 @@
+#!/bin/sh
+# IT CAN TAKE ONLY ONE FLAG
+# example:  " dcf h -33 <command> " or " dcf c -n <command>"
+
+one="$1" ; two="$2" ; three="$3" ; four="$4"
+
+#If no function is used, work as "which"
+[ -z "$two" ] && which $one && break && exit
+
+#Shift values, if there is no flag
+[ -z "$three" ] && three="$two" && two=""
+
+# Tells you which package owns that command,
+p() { pacman "${two:--Qo}" $( which "$three" ) ; }
+# or give it a file directly (for lib or share, which aren't executable)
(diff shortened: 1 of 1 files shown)


narukeh:
pro-tip: alias "dcf v" to "dcfv" to immediately open a script in vim (and others if you like). this makes it more comfortable to write.

I did this after seeing your shell scripting lmc YT video. I used to type the  pacman version by hand, when i was trying to remove bloat. So i thought i'd share that.

kronikpillow:
find your script? voidrice scripts are in path, you don't need to find them to run then, you just type the script name and it works

narukeh:
@kronikpillow Obviously. this is not to run them, this is to edit them (without writing its absolute path, and not limited to just larbs or local scripts). please read the script itself, im not very good at explaining. 😄

PR #521, opened 2020-03-28
Add dunst pywal/xresources theming script

drake-newell:
I found and adapted a script that generates a dunstrc color scheme based on xresources/pywal colors. It will replace/overwrite the old dunstrc with the new color scheme every time the script itself or pywal/setbg is run.


LukeSmithxyz:
Wait, so what does this add that the preexisting dunstrc wal template doesn't have?

drake-newell:
@LukeSmithxyz I was not aware of this dunstrc wal template. Can it generate colors from regular xresources as well?

LukeSmithxyz:
Well I think by default it would just use the most recent wal scheme.
I'm trying to think of a more efficient way to do that.

drake-newell:
@LukeSmithxyz where could I find this or is it in the voidrice dunstrc?

kronikpillow:
@LukeSmithxyz I wanted to open an issue about this, I find it very annoying that if pywal is installed it automatically loads the pywal color scheme, in my opinion, it should be toggleable and also if pywal was run once dunstrc gets changed and you basically can't change it back (or I couldn't find a way), I actually love the default color scheme shipped with voidrice, but sometimes when I just want a temporary "bling" to my desktop/terminal I would like to have pywal installed and to use it with the setbg script.
another annoying part is that your dwmbuild loads the pywal colorscheme by default if pywal is installed

@drake-newell it's in the setbg script and .config/wal

drake-newell:
@kronikpillow the default version of this script made a specially named dunstrc that wouldn't overwrite anything

drake-newell:
@LukeSmithxyz I think the benefit of this script is that it allows a dunst color scheme from xresources as well.

drake-newell:
@LukeSmithxyz I think the benefit of this script is that it allows a dunst color scheme from xresources as well.

PR #534, opened 2020-04-02
sxiv: Prevent writing all marked imgs onto file (lossage)

equwal:
Running `C-x c` `copy` or `C-x m` `move` will copy/move all the marked files to a
single location. If that location is not a directory, then creating that file
and overwriting it several times results in complete loss of data in the case of
moving, and unhelpful behaviour in the case of "copying". Only the last marked
file remains in the file.

This fix only moves/copies the first marked file, then issues a warning message.

sxiv-script-warnage
(My StumpWM gentoo linux system, and a Brazilian cat)

P.S. Send me pictures of your cat if reading this, thank you.


LukeSmithxyz:
Do you really need the `MULTI` variable? Couldn't you just do a directory check regardless and break the while loop?

equwal:
> Do you really need the `MULTI` variable? Couldn't you just do a directory check regardless and break the while loop?

@LukeSmithxyz 
Indeed, that is the case. I also think it is probably better to do nothing than move the first mark (which is almost never what you want). It looks ready to go now.

PR #538, opened 2020-04-04
replace hardcoded paths with $XDG_*_HOME variables, and clean up a few more programs

dennisleexyz:
This PR does 2 things:
1. Replaces instances of `~/.config`, `$HOME/.config`, `~/.local/share`, and `$HOME/.local/share` with `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME` appropriately. This should allow the user to move those locations anywhere and have everything just work. I wasn't sure whether to apply the changes to the documentation files as well; I ended up doing so, but that can be reverted. Note: may require #526 to be merged first to avoid breakage.
2. Adds a few more clean-up rules to `.profile` for a few different programs (`tmux`, android-tools, and rust's `cargo`) to get them out of `~/`.


LukeSmithxyz:
Great. I had been meaning to get to this... Could you revert the changes in the readme? I want as minimal confusion as possible for new users, while people who are going to change these variables know what they're doing.

dennisleexyz:
OK, I reverted both the repo readme and the `Mod+F1` one. I assume that's what you meant.

.local/share/larbs/getkeys/sxiv
```diff
@@ -12,4 +12,4 @@ sxiv is the image viewer.
 	r - Reload image if changed
 	m - Mark/unmark image
 	w - Zoom to fit window
-	ctrl-x - Run external command (see ~/.config/sxiv/exec/key-handler for options)
+	ctrl-x - Run external command (see $XDG_CONFIG_HOME/sxiv/exec/key-handler for options)
```
aldenjenkins, line 15: Should this be verbose as its human facing?

dennisleexyz: Hmm, I missed that one. I'll remove it tomorrow.

dennisleexyz:
Should be ready to merge now. I resolved the conflict with the new changes to the weather script, and even tacked on getting rid of `~/.mbsync`.

LukeSmithxyz:
No, don't get rid of `.mbsyncrc`. mbsync is called both in the shell and programatically in mutt-wizard, so if I have an alias line like that, I also must force the different location in mutt-wizard, which causes confusion because people who run mutt-wizard without the alias will be running two different "config files" whether it is run manually or in the script.

We'll have to wait for isync to actually allow a better config location.

dennisleexyz:
Done.

LukeSmithxyz:
Alright. I've merged this with the additional change of setting a default value in the scripts so that they will still function if independently downloaded and used on systems without xdg environmental variables.

jlaw:
Why do you need to set default values in all the scripts if the XDG variables are already set explicitly in .profile by reading user-dirs.dirs?

The unintentional effect of this merge is init.vim doesn't load plugins properly as it takes $XDG_CONFIG_HOME as literal.

I also believe that the proper way to use a default with undefined variables is by adding a '-' after the colon:
${XDG_CONFIG_HOME:-$HOME/.config}

LukeSmithxyz:
Yeah I guess I missed the `-` in my sed commands, it's been put there now.

> Why do you need to set default values in all the scripts if the XDG variables are already set explicitly in .profile by reading user-dirs.dirs?

For the reason I just said above...

NikkiSatmaka:
I still don't understand how this xdg works. The lf config doesn't work
I've read the https://wiki.archlinux.org/index.php/XDG_user_directories

I've also tested on creating a new user using larbs, but it doesn't work out of the box
I've added user-dirs.dirs di XDG_CONFIG_HOME path, but it doesn't change anything

Would anyone care to share some references or examples on to make this work?

jlaw:
According to [this ArchWiki page](https://wiki.archlinux.org/index.php/XDG_Base_Directory), only `$XDG_RUNTIME_DIR` is set by default through `pam_systemd`; all other variables need to be explicitly defined by user.

I'm not sure why @LukeSmithxyz is using the file `$XDG_CONFIG_HOME/user-dirs.dirs` to keep these variables because, if a LARBS user decide to install [xdg-user-dirs](https://www.archlinux.org/packages/extra/x86_64/xdg-user-dirs/), the file `$XDG_CONFIG_HOME/user-dirs.dirs` is overwritten, removing variables `$XDG_CONFIG_HOME`, `$XDG_CACHE_HOME`, and `$XDG_DATA_HOME`, since they are not recognized by `xdg-user-dirs` according to the [user-dirs.dirs man page](http://man.he.net/?topic=user-dirs.dirs&section=all).

Regardless, the magic for setting the defaults in LARBS happens at [these lines from .profile](https://github.com/LukeSmithxyz/voidrice/blob/a8255f3b478638b91d60ba57c8cf030ad121193a/.profile#L17-L18).

LukeSmithxyz:
> I've also tested on creating a new user using larbs, but it doesn't work out of the box
> I've added user-dirs.dirs di XDG_CONFIG_HOME path, but it doesn't change anything
> 
> Would anyone care to share some references or examples on to make this work?

What do you mean by "doesn't work"? What are you trying to do?

NikkiSatmaka:
> > I've also tested on creating a new user using larbs, but it doesn't work out of the box
> > I've added user-dirs.dirs di XDG_CONFIG_HOME path, but it doesn't change anything
> > Would anyone care to share some references or examples on to make this work?
> 
> What do you mean by "doesn't work"? What are you trying to do?

I just opened lf after a fresh LARBS install, and the config, like icons, don't show up properly. If I change the XDG back to ~ the icons would show up.

asakura42:
> No, don't get rid of `.mbsyncrc`. mbsync is called both in the shell and programatically in mutt-wizard, so if I have an alias line like that, I also must force the different location in mutt-wizard, which causes confusion because people who run mutt-wizard without the alias will be running two different "config files" whether it is run manually or in the script.
> 
> We'll have to wait for isync to actually allow a better config location.

But mutt-wizard reads $MBSYNCRC variable, doesn't it? Maybe give that alias a chance and put `mbsync="mbsync -c ${XDG_CONFIG_HOME:-$HOME/.config}/isync/mbsyncrc"` in aliasrc and `export MBSYNCRC="${XDG_CONFIG_HOME:-$HOME/.config}/isync/mbsyncrc"` in .profile?

PR #540, opened 2020-04-06
New feature: Zsh named dirs

karioki:
I think this is useful. It uses the same directory shortcut file (~/.config/directories)

You can do things like this from Zsh shell:
`$ mv file.tex ~d`
Or even better and more relevant example:
`$ rsync ~d/file.tex ~D/project.tar user@server:~/`

The main idea is the user doesn't have to write the entire path when using Zsh shell only.
They can use any binding from the directory shortcut file.

Filename completion works too when using this.


LukeSmithxyz:
Cool idea!

karioki:
> Cool idea!

Thank you!

PR #542, opened 2020-04-06
wttr options

jonmed:
This adds options to the wttr link that basically remove bloat that the module does not need, making parsing the file easier.
The options are listed [here](https://wttr.in/:help). 

The ones added here are:
1 -- only get the current day report
F -- do not show the "Follow" line
Q -- superquiet version (no "Weather report", no city name)
T -- switch terminal sequences off (no colors)


LukeSmithxyz:
This is all intentional. There is an alias `weath` that shows the entire downloaded weather report in color if the user likes. I think in i3blocks it comes up when clicked.

PR #545, opened 2020-04-08
Move bg link to ~/.local/share/

narukeh:
often the ~/.cache directory is deleted because it gets too big, in my case if i compile stuff with yay, firefoxs cache also gets quite big, and mesa shaders are also stored there.

### DIFF
--- a/.cache/bg
+++ b/.cache/bg
@@ -1,1 +1,0 @@
-../.config/wall.png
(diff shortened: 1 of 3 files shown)


LukeSmithxyz:
Okay, good suggestion then. Could you also add pre-existing a link in that location to the old wallpaper like I have for the current location in this same PR?

narukeh:
@LukeSmithxyz i'have done that, and also removed the .cache directory.

PR #557, opened 2020-04-11
i3/config volume update for statusbar

narukeh:
after lmc removal, forgot to add `pkill -RTMIN+10 i3blocks`


LukeSmithxyz:
Now the dwm version was lagging on multiple volume changes, so I have it just update every second instead. I'm going to go ahead and merge this for the time being and see if the same problem is present on i3. Otherwise, later I'll remove this and change the i3blocks config.

PR #559, opened 2020-04-11
fix: Restore Mod + F2 functionality of dwm

jlaw:


LukeSmithxyz:
Get the most recent build of my dwm. This is not needed anymore.

jlaw:
Ah OK, I was using `Mod+F2`; didn't see the command changed to `Mod+R`. Perhaps LARBS help will need some updating after all your refactoring is done.

drake-newell:
@LukeSmithxyz What's the benefit of doing it this way?

LukeSmithxyz:
So the patch I installed allows signalling to dwm to refresh or quit.

I wanted the ability to "refresh," esecially to redraw the bar and other
things new in case of, let's say, monitor changes (plugging a laptop
into a dock). So now, I can have an external script tell dwm to restart
when needed.

Additionally, I can signal to dwm directly to quit from a shell, so now
to quit, I don't do something potentially suicidal like killall Xorg
(potentially killing your i3 instance on another terminal), but just
signal to the found dwm instance to terminate.

I couldn't just killall dwm, beacuse that would just restart it in the
past.

So now that this feature exists, I might as well use it than the silly
hackish loop to restarting.

Additionally, I was toying with sx as a replacement of xinit and the
loop doesn't even work on it, so this works better for that too.

PR #562, opened 2020-04-12
tell Qt to use whatever the current GTK2 theme is

dennisleexyz:
Just this one change for now, but there's a couple more that I can add if you're interested in upstreaming them:
* (in `~/.config/files`) alias cfp -> `~/.profile`
* (in `xprofile`) set an environment variable to make touchscreen input work correctly on Firefox and Firefox-based browsers

### DIFF
--- a/.config/xprofile
+++ b/.config/xprofile
@@ -19,3 +19,6 @@ xcompmgr &		# xcompmgr for transparency
 dunst &			# dunst for notifications
 xset r rate 300 50 &	# Speed xrate up
 unclutter &		# Remove mouse when idle
+
+# program settings:
+export QT_QPA_PLATFORMTHEME="gtk2"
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
I think I'd rather have this in the profile file with all the other environmental variables. I mean sure it's a graphical thing, but I don't think it would harm anything including it in the general profile.

What is this variable to do with Firefox?

dennisleexyz:
[The Firefox thing](https://wiki.gentoo.org/wiki/Firefox#Enabling_multitouch) is `MOZ_USE_XINPUT2="1"`; it makes scrolling smooth like it is on Chromium-based browsers instead of quantized like it is regularly and makes touchscreens work as you would expect instead of treating it like a mouse click. I have it on my personal setup.

PR #565, opened 2020-04-13
New Ergonomic Mapping

karioki:
Try this out before merging.  It's very different from the usual mapping.

It's not perfect but I've been using it now for a long time.

So it modifies some keys:
Key:-              SinglePress     Hold 
SpaceBar      Space               Super or Window Key
CapsLock      ESC                  Ctrl
Enter             Enter                Ctrl
LeftShift        Underscore     LeftShift
RightShift     Dash                 RightShift

So in a short summary, this will give Control keys access to the pinkies in addition to ESC.  
And will use two Shift keys for frequently used but hard to reach keys (at least for me).
Last but not least the SpaceBar act as Super key which can be super useful.

Some problems:
- 'super + space' will not work from the dwm so I am using 'super + w' instead
- Very rarely, in vim when you type <Leader><key> this can go as Super + <key>. This is mainly due to slow xcape or human error.

### DIFF
--- a/.local/bin/remaps
+++ b/.local/bin/remaps
@@ -3,9 +3,40 @@
 # This script is called on startup to remap keys.
 # Increase key speed via a rate change
 xset r rate 300 50
-# Map the caps lock key to super...
-setxkbmap -option caps:super
-# But when it is pressed only once, treat it as escape.
-killall xcape 2>/dev/null ; xcape -e 'Super_L=Escape'
-# Map the menu button to right super as well.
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Well that's an interesting setup, I've been suggested to use space for super before, but I don't think I'll want to merge this.

To be honest, I'm thinking about peeling back some of the keyboard problems because they cause real trouble and confusion on foreign keyboards and such. I think this setup is something else very liable for that kind of breakage.

PR #578, opened 2020-04-18
hide go folder

mighty9245:
go folder is the go language workspace and is placed in your home directory by default. even if you don't use go, it will be created in process of building something written in go. this moves go folder to .local/share/go.

i'm not writing in go, maybe there is a point of keeping this folder in your home directory, but i don't really see it.


tallguyjenks:
thanks for this, i've been wanting to get that out of there!

Issue #577, opened 2020-04-17
dmenupass - password not hidden


MarcBeaudet:
I seem to have a problem with the dmenupass script. 
The password is not hidden. 
The same happens with sudo -A.

No modifications were made on the script. Should it require specific settings/programs?

Thanks

drake-newell:
I would suggest that Luke add the password patch to his dmenu build

MarcBeaudet:
Thank you, using the password patch + adding the `-P` flag in dmenupass fixed my issue.

PR #588, opened 2020-04-21
dwm compliant

charlie39:
revised PR
simply displaying the current song playing [or paused] in  status bar in dwm.

### DIFF
--- a/.local/bin/statusbar/music
+++ b/.local/bin/statusbar/music
@@ -1,7 +1,10 @@
 #!/bin/sh
 
+dwm="$(pgrep 'dwmblocks')"
+
 filter() {
-	sed "/^volume:/d" | tac | sed -e "s/\\&/&amp;/g;s/\\[paused\\].*/<span color=\"gray\" font_style=\"italic\">/g;s/\\[playing\\].*/<span>/g" | tr -d '\n' | sed -e "s/$/<\\/span>\n/g"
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Use pidof instead of pgrep. Much, much faster. I've been trying to switch all my pgreps over.

charlie39:
Okay dokay, I will do that .

I am going through the repo and saw some overhaul since the time I used. 
I have some questions now, 

1. where is i3blocks/dwmblocks is initialized since it's separate for both now?

2. Where is mpd-module-update script used since I don't see it in xinitrc ? this PR is dependent on that no?

3. $ALSA_CONFIG_PATH isn't really helpful.
if one uses that and moves the ~/.asoundrc there, then aplay -l throws errors. I think I will open an issue on this.

charlie39:
I mean in the xprofile not in xinitrc.
I see this, 
https://github.com/LukeSmithxyz/voidrice/issues/577#issue-602282674

LukeSmithxyz:
Yeah I've noticed the ALSA thing doesn't work as expected. The dotfiles don't use ALSA anymore anyway though. Or at least they use Pulseaudio.

The music statusbar module is not active by default now due to some problems I hadn't bothered to solve. I think a recent PR might've fixed them though.

i3blocks is started by i3. dwmblocks is started by dwm.

PR #605, opened 2020-04-28
Upstream problem with java not able to figure dwm's windows.

charlie39:
Jetbrains products (intellij Idea and pycharm) and Netbeans draws grey windows in dwm(upstream bug). 
Intellij Idea will work with this two env vars. for pycharm, affected plebs may need to install the wmname tool (community repository)


vladdoster:
Or you could just use nvim with a coc language server 😬🤔👀

charlie39:
@vladdoster thanks man, I have set with ALE and I try do as much as I could in nvim. But when it comes JAVA EE(web application,) there's no option but to use one of those IDE.

LukeSmithxyz:
I've allowed these to be added now. I'm going to have to have a more elegant way of dealing with environmental variables if people keep adding them for obscure programs.

PR #606, opened 2020-04-28
Remove *very* dangerous code

MinestoPix:
.bak stands for _backup_. A backup should **never** be automatically removed by any script.


LukeSmithxyz:
LaTeX automatically produces and overwrites bak files of the same basenames all the time. That's all this script removes. If you manually make you own backup files, just don't name after a .tex file and you'll be fine. They'd get removed with or without this script.

PR #607, opened 2020-04-29
Changes to mpd.conf to support pulseaudio

AlexandrosVasileiou:
After pulling the new version of LARBS, I could no longer listen to music. I noticed that Luke moved many things to Pulseaudio but mpd.conf was left behind. This fixed the music issue for me.

Feel free to implement it in a better way if there is :)

### DIFF
--- a/.config/mpd/mpd.conf
+++ b/.config/mpd/mpd.conf
@@ -9,6 +9,8 @@ max_output_buffer_size "16384"
 audio_output {
       type  "alsa"
       name  "ALSA"
+    # type "pulse" # uncomment for pulseaudio
+    # name "pulse" # uncomment for pulseaudio
 }
 
 audio_output {


LukeSmithxyz:
I hadn't switched this over because it totally breaks the audio setup on my machine as mpd and pulseaudio fight for the same socket I think.

What distro are you using and how do you start mpd? With a system service? By a particular user?

AlexandrosVasileiou:
I've been using Arch with Larbs for about a month. I have not discovered Larbs in depth yet, but I can tell you that I start mpd in the default Larbs way, which (if I'm not mistaken) is done via the .xprofile which gets sourced by xinit.

I may have tweaked something by mistake and ALSA didn't work for me, as I installed Larbs on a new user and it works with the ALSA setting for mpd.

A hint: For my main user, (the one which I need pulse in mpd to work) I get the following error when running `alsamixer` on the terminal:
```
ALSA lib control.c:1379:(snd_ctl_open_noupdate) Invalid CTL default
cannot open mixer: No such file or directory
```

LukeSmithxyz:
I think I'll go ahead and merge this since it's just comments and do more investigation later on Arch proper when I get the chance.

Issue #616, opened 2020-05-06
dwm fresh install weather report doesnt display correctly


tallguyjenks:
lookin through past issues #509 resolved the issue, and on current install less's args are `-Sf` and what resolved was using the `-r` option in #509 not sure if it was changed for some reason but on a brand new system and fresh dwm install weather report doesnt display correctly without `-Srf`

LukeSmithxyz:
What's your `LANG` variable? It works fine on mine. The reason I changed it now is that it shouldn't wrap lines when in a small terminal.

tallguyjenks:
`LANG` is `en_US.UTF-8`

LukeSmithxyz:
Jeez, here we go again... I never figured out why it doesn't work on anyone but myself's computer lol.

Tell me, if you spawn the preview (by clicking the statusbar) when there are other windows, does the weather report wrap the text (like if the window is too small) or does it not show the overflow text? Does is make a difference with/without `-r`?

tallguyjenks:
Here are the two cases with and without the `-r`option

With `-r`

![with_r_option](https://user-images.githubusercontent.com/29872822/81221147-49d8cc80-8f97-11ea-8789-c1d619b93459.png)

without `-r`

![without_r_option](https://user-images.githubusercontent.com/29872822/81221152-4ba29000-8f97-11ea-9cfa-9c07297969bf.png)

tallguyjenks:
i dunno man, im running on x220's as well, ¯\_(ツ)_/¯

LukeSmithxyz:
What is your `LESS` value equal to? Is it `-R`?

tallguyjenks:
@LukeSmithxyz yup, that was the default `echo $LESS` = `-R`

BachoSeven:
Since I don't feel like it's worth opening a new issue, I want  to report that this also applies to the new `price` script: I have LESS set to "-R", lesspipe.sh installed and configure but unless I explicitly use `less -r` in the script it won't display raw chars.

narukeh:
not dwm specific. im still using i3.
with `-r` which is the defaulf, so when i click i get:
![with-r](https://user-images.githubusercontent.com/62762955/83333100-23106c00-a29f-11ea-9ee3-8f809b24b371.png)

without `-r`
![without-r](https://user-images.githubusercontent.com/62762955/83333108-34597880-a29f-11ea-991f-005a273b1cc5.png)

```
$ echo $LESS
-R
```

[...]

PR #622, opened 2020-05-09
without -e, the ansi escape codes dont function

tallguyjenks:


LukeSmithxyz:
Well:

1) This script is obsolete since the individual statusbar modules are clickable now, so I'll probably just delete this file now that you remind me of it.
2) echo flags are not POSIX compliant and this would break the script where `dash` or something else is `/bin/sh`.
3) It works fine as of now on my machine. If it's not displaying for you, I suspect it's the same issue as  https://github.com/LukeSmithxyz/voidrice/issues/616.

PR #623, opened 2020-05-09
Essential alias: ll

karioki:
I think people are used to it so we should add it.


exlumine:
would also add `l` for regular `ls`, and `la` as regular `ls -a...`

LukeSmithxyz:
As a rule, I add default features with aliases and I abbreviate commands that are too many letters: `killall`, `transmission-remote`, `youtube-dl`.

I don't like adding aliases for specific options (unless I want those options to be run all the time) because this basically opens up the door for any number of arbitrary aliases for random use cases.

PR #625, opened 2020-05-09
Replace tr(1)/sed(1) combo with paste(1)

kdkasad:
The `paste(1)` command is designed to concatenate lines of input. Several scripts used something like `tr '\n' '+' | sed 's/+$//'`. I've replaced this with a much simpler `paste -sd '+'`.


LukeSmithxyz:
Oh cool. I didn't know of this command.

PR #626, opened 2020-05-10
should be KiB not kB and more precise RAM measurement

vladdoster:
if you're dividing by 1024 its a KiB not kB

https://en.wikipedia.org/wiki/Kibibyte


vladdoster:
smh git

tallguyjenks:
holy shit dude xD

PR #629, opened 2020-05-11
Update xprofile to fix some Java Applications that don't render correctly in DWM.

cm-mcl:
Add Environment Variable to fix Java Application Rendering in DWM, doesn't cause any issues in i3 from what I've seen. Feel free to check yourself, though.


LukeSmithxyz:
Okay. I'm going to add this to the profile file though as that's where I tend to put environmental variables (even graphical ones).

PR #631, opened 2020-05-12
Add tmux conf to file shortcuts

vladdoster:


LukeSmithxyz:
I'm sort of thinking of removing the tmux config because I don't really use it and it's only used for i3, but okay.

PR #634, opened 2020-05-14
Add keyboard layout related scripts

MYDavoodeh:
`kblayoutmanager` switches layouts.
Layouts are read from `$KBLAYOUTS`.
An extra file, `~/.local/bin/statusbar/lang` is added in the commit,
which adds layout indicator to i3blocks.

Example binds are `Super+Alt+Space` added to both i3/config and
sxhkd/config.
Example `KBLAYOUTS` are US Standard layout and Global UK layout.

Special thanks to KEZ-

I proposed this since I was unable to push on the previously opened issue #472 

If you want to apply this right now run the following commands:
```sh
curl -OL https://patch-diff.githubusercontent.com/raw/LukeSmithxyz/voidrice/pull/634.patch
git am < 634.patch
```

### DIFF
--- a/.config/i3/config
+++ b/.config/i3/config
@@ -46,6 +46,7 @@ bar {
 bindsym $mod+Return 		exec $term
 bindsym $mod+Shift+Return	exec --no-startup-id samedir
 
+bindsym $mod+Mod1+space 	exec --no-startup-id kblayoutmanager cycle $(echo $KBLAYOUTS)
 bindsym $mod+Shift+space 	floating toggle
 bindsym $mod+space		focus mode_toggle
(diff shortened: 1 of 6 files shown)


MYDavoodeh:
`.local/bin/kbselect` makes this obsolete and you can simply add it to your blocks by inserting this in your config:
```
[kbselect]
interval=once
signal=30
```

A compatible script with that module (for binding and not "dmenu select") is located in my repo in here: https://raw.githubusercontent.com/MYDavoodeh/dotfiles/renew/.local/bin/kblayout


I'm closing this PR.

PR #635, opened 2020-05-16
keyboard layout scripts (dwmblocks + dmenu)

BertramMiller:
These scripts work with dwmblocks and allow for easy switching between keyboard layouts

### DIFF
--- a/.local/bin/langselect
+++ b/.local/bin/langselect
@@ -1,7 +1,6 @@
 #!/bin/sh
 # requirements: dmenu, xorg-setxkbmap
-choices="us\\nfr\\nua\\nru"
-chosen=$(echo -e "$choices" | dmenu -i )
+chosen=$(localectl list-x11-keymap-layouts | dmenu -i )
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
Well, I want a feature like this, but ideally I'd like to not have to hardcode the alternate languages/keyboard. I'd prefer either:

1) Getting the desired default non-English keyboard by looking at system settings. For example, some systems use `/etc/vconsole.conf` to key the default keyboard. There might be other places where this is kept or more universal, but I don't know since I just use an English keyboard for most things.

2) Just giving a dmenu list of all keymaps or at least the most common ones.

BertramMiller:
Alright, here is the simple solution, it covers 99 layouts:

localectl list-x11-keymap-layouts | dmenu -i

Some layouts, such as Traditional Chinese would require additional setup, installation of fonts etc.

LukeSmithxyz:
Well... `locactl` is systemd specific. I run Artix myself and I would like it to work on any init system if possible. I'd have to look up a runit and/or openrc equivalent if any...

PR #642, opened 2020-05-18
Clean up of .zprofile, any script using an XDG env var, and X11 .xinitrc/Xresources

vladdoster:
I don't get why there is parameter expansion everywhere? Wouldn't every XDG var be set if they are using `voidrice`? Seems like a lot of extra stuff for not alot of benefit?

You can just do:
```
$ export INPUTRC="$XDG_CONFIG_HOME"/readline/inputrc
```
This PR does the following:
- Condenses X11 resources into their own folder (Do I remember it being this way at one point?)
- Condenses inputrc into readline dir which can be found [here](https://wiki.archlinux.org/index.php/XDG_Base_Directory#Supported
)
- Cleans up redundant env variable cruft
- Update any files to reflect changes
- Update i3 things that were referencing the old xprofile in .config
- Fix bashisms such as ~/ instead of using ${XDG_RUNTIME_DIR}


LukeSmithxyz:
Why are you trying to load the xprofile in `XDG_RUNTIME_DIR`? That's basically a temp directory obviously out of this repository.

vladdoster:
> Why are you trying to load the xprofile in `XDG_RUNTIME_DIR`? That's basically a temp directory obviously out of this repository.

I misinterpreted what it did. I thought it was the $HOME dir for some reason. Re-reading the Arch wiki and it is clear that is not what it does. I changed all instances to `$HOME`
```
grep -rl {XDG_RUNTIME_DIR} . | xargs sed -i 's/{XDG_RUNTIME_DIR}/$HOMEt/g'
```

LukeSmithxyz:
I don't know. Merging this seems like a huge inconvenience for me. It basically means that people can no longer just take scripts from my repository and run them. I have to specifically tell them (every single time) that they have to have XDG variables active in their shell environment.

Now the profile variables should assign fine. I could make these changes to `~/.zprofile`. The issue is that in the shell scripts cannot use XDG variables unless they have been specifically exported. Obviously they are in my dotfiles, but it's a pain to everyone else.

If I'm going to make these changes, I don't know why I shouldn't just hardcode each as their default values, i.e. just `~/.config/` instead of the uglier `$XDG_CONFIG_HOME/`. The first would actually work on any system. The second, only on those where the variable is set.

vladdoster:
I have acctually changed my mind about this hahaha. I ran into the exact problem at work when I wasn't to transfer my neovim file and the XDG_CONFIG_HOME wasn't set so it broke neovim. I'm closing this and glad I experienced it to know. The way it is now is honestly the most portable, even though verbose and ugly.

PR #643, opened 2020-05-20
stderr suppression & empty string flag

anntnzrb:
- `[ "$archive" = "" ] ` ==` [ -z "$archive" ]`
- upon entering `ext -c` with no arguments you will also generate an error from `readlink`, should be suppressed

also sorry for the formatting, `shfmt` got in the way

### DIFF
--- a/.local/bin/ext
+++ b/.local/bin/ext
@@ -7,38 +7,38 @@
 # Behavior with `-c` option: Extract contents into current directory
 
 while getopts "hc" o; do case "${o}" in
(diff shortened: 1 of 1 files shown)


vladdoster:
@nntnzrb, just a tip, I've been told off plenty of times for it. Don't mess with formatting, especially if it isn't your codebase. You want your diffs to be as precise as possible and to encompass the actual changes you make. It's not only hard to decipher your changes, but it doesn't look as good. Not a problem but it already makes a PR need some touch-up.

PR #653, opened 2020-05-27
Copy audio codec instead of encoding audio

rickyjon:
Also for the last command, the -ss option is behind of -i, it should seek faster compared to "... -i -ss ..."


vladdoster:
Huh, I was always under the assumption order of flags didn't matter.

iotku:
As far as I'm aware it shouldn't make a difference, but ffmpeg can be quirky like that.

Anyways, I made a similar pull request here https://github.com/LukeSmithxyz/voidrice/pull/654 which should cover some extra edge cases around file extensions as currently the script is hard coded to the "opus" extension, which will generally cause failures should the input file not be opus.

rickyjon:
> Huh, I was always under the assumption order of flags didn't matter. Since the way it's parsed is obfuscated since it's a binary?
Order of flags sort of matter, at least the position around the -i flag, for example:
"ffmpeg -ss 05 -i <file> -ss 05 -t 5 ..." is the same (output-wise) as "ffmpeg -i <file> -ss 10 -t 5 ...".

LukeSmithxyz:
Wouldn't this cause problems if the target container is different from the input container?

vladdoster:
You learn something new everyday. Thank you @rickyjon

iotku:
> 
> 
> Wouldn't this cause problems if the target container is different from the input container?

I addressed this here https://github.com/LukeSmithxyz/voidrice/pull/654 by changing the extension variable to `ext="${1##*.}"` (Is this a bashism™?) which would maintain the original file extension (and thus stream copy would work fine), the issue being that the tag script doesn't support the format I was using for testing.

So of course my next logical conclusion was to have ffmpeg do the tagging itself (after all I wasn't happy about the opus tag dependency which I had to manually compile under Debian because I didn't see it in the repos which was non-trivial).

Sounds great right? It is, except there doesn't appear to be any consistent tagging between different formats so that would require extra logic for different file types. 

For example, `-metadata track="1"` was ending up in the **track number** field for aac (m4a), but under the **disk number** section for opus (webm). And `-metadata artist="Artist"` was ending up as **Artist**, for m4a and **Album Artist** for opus.

It's not the end of the world, but it's annoying.

rickyjon:
> That home folder though in your dotfiles

@vladdoster, Not even mine


> Wouldn't this cause problems if the target container is different from the input container?

@LukeSmithxyz, It depends on the codec and the container (for example .ogg and .mkv, can use the same opus audio codec but opus in .mp4 is rejected by ffmpeg by default), since the script always wants it into .opus, it can only realistically accept opus codec, The way I see it is that by constantly seeking re-encoding every Nth amount of time seems wasteful instead of re-encoding input with the right codec correctly. There's also the issue of dealing with date time in POSIX Shell being very tricky without using better programs, so if you were to re-encode and make it noticeably faster you would have to try to convert the timestamps into seconds, so you can have "ffmpeg -i -ss (timestamp) -t (elapsed amount of seconds between the first timestamp and the second) ...".
If you really want to just throw whatever and not worry about that, just create an ffprobe check + case statement to figure out what audio codec it is and use the right container for it and probably make use of multi-threading/doing all the encoding at the same time OR just use a container that supports many audio codecs like mkv for example.

vladdoster:
@rickyjon all in jest :)

PR #654, opened 2020-05-28
[Booksplit] Remux / Remove opustag dependency

iotku:
By using the original file's extension and ffmpeg's stream copy functionality the script is greatly sped up while avoiding losing quality to transcoding audio.

speedup

Additionally, as ffmpeg already has tagging functionality, it's pointless to roll our own so we let ffmpeg handle the tagging instead.


iotku:
Eh, self tagging with ffmpeg might add a bunch of nasty logic (for instance track number seems to go into disk number depending on the codec) so I'm going to sit this one out for now.

PR #658, opened 2020-05-28
Use prompt argument in read

breadcat:


LukeSmithxyz:
`-p` is not POSIX compliant.

PR #659, opened 2020-05-28
replaced a bashism with posix equivalent

barudaret:
It's a tiny change, but this will let the script work on systems like mine where /bin/sh is symlinked to dash rather than bash.  Maybe you'd prefer expr instead of bc, let me know


barudaret:
Actually nevermind.  This is a non-issue.  I could'a sworn double-parens is a bashism, but apparently dash can cope with it

PR #664, opened 2020-05-30
Use read instead of parameter expansions

MaanooAk:
A suggestion to get rid of the parameter expansions (I always found them weird). 

`read` could also be easily changed to handle more than two values (if someone would like to keep track of more statistics).


LukeSmithxyz:
I actually already commited something like this locally and just haven't pushed everything yet.

PR #665, opened 2020-05-30
Replace case with a function call

MaanooAk:
The idea: Replace
```bash
case "$1" in
	toggle) toggle ;;
	mute) mute ;;
	up) up ;;
	down) down ;;
	control) control ;;
esac
```
with
```bash
"$1"
```

But then you can also do `lmc ls` and the ls will run, so the `type "$COMMAND" | grep function` ensures that the passed command is one of the commands.

The `help()` is just bloat.

### DIFF
--- a/.local/bin/lmc
+++ b/.local/bin/lmc
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+COMMAND="${1:-help}"
 NUM="${2:-5}"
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Oh I forgot this script is still in the repository. I meant to delete it.

PR #674, opened 2020-05-31
Refactor nettraf

chambln:
Here, a  function `update` is used to factor out common I/O and arithmetic. Summation is carried out by a simple _for_ loop instead of **paste**(1p).

### DIFF
--- a/.local/bin/statusbar/nettraf
+++ b/.local/bin/statusbar/nettraf
@@ -14,12 +14,12 @@ update() {
     sum=0
     for arg; do
         read -r i < "$arg"
-        sum=$(( "$sum" + "$i" ))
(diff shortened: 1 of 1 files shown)


narukeh:
why did u remove the quotes from ` case "$BLOCK_BUTTON"` tho?

chambln:
> why did u remove the quotes from ` case "$BLOCK_BUTTON"` tho?

The variable of interest in a case statement is not subject to field
splitting, so the double quotes are not necessary (although they do no
harm really).

To convince yourself of this, try running

``` bash
#!/bin/sh
IFS=' '
spacious='foo bar'
case $spacious in 'foo bar') echo still works ;; esac
```

and you should see 'still works' echoed out.

PR #676, opened 2020-06-01
Shellcheck

WhispersOfJ:


.local/bin/qndl
```diff
@@ -6,7 +6,7 @@ base="$(basename "$1")"
 notify-send "⏳ Queuing $base..."
 cmd="$2"
 [ -z "$cmd" ] && cmd="youtube-dl --add-metadata"
-idnum="$(tsp $cmd "$1")"
+idnum="$(tsp "$cmd" "$1")"
```
cupnoodles14, line 9: this seem like it would break stuff when $cmd has args and splitting is actually desired (like youtube-dl line above).

same for chunks with $TERMINAL and $BROWSER (if someone sets them to contain args, which is pretty rare, but IMO it's even more rare for them to have spaces that need escaping)

PR #677, opened 2020-06-01
cleaning

narukeh:
making it a variable since its used 5 times


narukeh:
NOTE: the `-r` should be removed from `less` since it makes `-S` ineffective.
(ill do that in another commit if you agree)

PR #691, opened 2020-06-05
Some annoying xdg base directory changes

salman-abedin:


salman-abedin:
It's a syncing tool.
It uses rsync to sync directories.
Super useful if you wanna do your syncing using command line.

LukeSmithxyz:
`$HISTFILE` is used by other shells (zsh) I believe, so it shouldn't be in a bash directory, but somewhere neutral. I'm fine with one dir level up.

salman-abedin:
How about in a directory named sh?
Cause the alternative is that it would be dangling on the share directory without any context.

LukeSmithxyz:
~/.local/share is fine. Don't assume that some other directory will be
created beforehand.
On 20/06/06 04:33PM, Salman Abedin wrote:
> How about in a directory named sh?
> Cause the alternative is that it would be dangling on the share directory without any context.
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/LukeSmithxyz/voidrice/pull/691#issuecomment-640132021

--
https://lukesmith.xyz

salman-abedin:
That makes better sense.

PR #692, opened 2020-06-06
Much cleaner compiler script

salman-abedin:


salman-abedin:
It's just the pandemic man! Extremely bored. Got nothing better to do but mess with lukes code. lol. XD

PR #693, opened 2020-06-06
Cleaner & bug free startx

salman-abedin:
My goal here was to clean up a bit and use one less program (pidof) to start the x server.
But after further testing, I have noticed this fixes a major issue.
For some weird reason, your version auto started the X on  all of my VTs instead of the first one.
It might just be the fact that I am auto logging on my VT, I'm not sure though.
Either way, my version seems to be working fine without any issue.


vladdoster:
👀👀

PR #700, opened 2020-06-08
Unmessy login screen

salman-abedin:


hamidout:
love this, much looks cleaner

salman-abedin:
Yeah, It looked like a redundancy to me but since I got this from the arch wiki, I figured I should do my homework (which I haven't yet, =) ) before dismissing it.

salman-abedin:
> And why exactly do you need the `-- vt1` thing when it already works?

Yeah, It looks like a redundancy to me but since I got this from the arch wiki, I figured I should do my homework (which I haven't yet done, XD ) before dismissing it.

narukeh:
@salman-abedin  i say `exec startx > $XDG_DATA_HOME/startx.log 2>&1`
It's still unmessy, plus you don't loose any info 😉

PR #701, opened 2020-06-08
make posix shell the default for shellcheck?

tallguyjenks:


LukeSmithxyz:
Well that would be counterproductive in bash and all the other languages it checks and unnecessary in shell. I mean if you have your script marked as `#!/bin/sh` it'll check for shell, not bash.

PR #703, opened 2020-06-09
This one needs bash to work where sh is not pointing to bash

karioki:
I guess, it was mistake from someone.


LukeSmithxyz:
Where is there a bashism?

karioki:
Somehow my merged code ended with the old code (with bash feature). Ignore it.

karioki:
I see now that bashism has been removed there ;)

PR #715, opened 2020-06-20
Chads urls file

narukeh:
UNREAD: lists everything that is unread (the articles will be removed from this "group" after reading, NOT deleted)
TECH: lists every feed that has the "tech" and that is unread (the articles will be removed from this "group" after reading, NOT deleted)
░░░░Luke Smith░░░░: serves a purpose of visual separation, but also not wasting a line, it will list everything that has the tag "luke" (read and unread)


AlexSyssoyev:
Hey this is amazing!! Is there a way to hide the normal rss feeds and just use the query's?

narukeh:
@AlexSyssoyev  nah, i dont think there is. but what youcan do is:
```
query: blah blah blah
query: blah blah blah
query: blah blah blah
" "
" "
" "
" "
```

AlexSyssoyev:
Thanks! i will try to set it up that way.

narukeh:
@AlexSyssoyev  i forgot that you actually can tag querys, then when you pres `t` it will show all the tags. so basically it will show u ur queries AND the tags.  🤷 not exactly what you asked but preeetty close.

AlexSyssoyev:
Hey i just saw in the [wiki](https://wiki.archlinux.org/index.php/Newsboat) that you can hide specific feeds. You just can add a ! before the tag or name.

SirCipherz:
The feed thing is awesome

PR #720, opened 2020-06-22
newsbloat 2.20 compliant

narukeh:
the `;` must be surrounded with space, or you wont be able to open stuff


narukeh:
btw, @LukeSmithxyz  you should re-ad nohup. without it, if you navigate in newsboat, while having mpv play a video (which came from newsboat), mpv will take that as commands, so it will change volume up and down, go backwards-forfards, depending on what you input while u are at newsboat.

PR #725, opened 2020-06-24
Make dir shortcuts use XDG DIRS and then use a default if need be

vladdoster:
Broke on my system so thought you might like it


vladdoster:
sweet! Forgot I had this open.

PR #726, opened 2020-06-25
Prevent any video lag when using camtoggle

aldenjenkins:
Lets help you prevent needing to sync audio after your vids!


vladdoster:
Regardless of the semantics, hardcoding isn't great to do. Maybe get max `resolution` and make that more dynamic. If I run my 4k webcam I don't want 720p as what's recorded. No?

PR #731, opened 2020-06-27
Prevent error when having no battery

Limero:
The current battery script works well on my laptop (that has a battery) but on my desktop (that lacks a battery), it floods my logs with the following because `cat` keeps throwing an error each time the script is run.
```cat: '/sys/class/power_supply/BAT?/capacity': No such file or directory```

This fix breaks the loop if the file doesn't exist before it's being read with `cat`, thus preventing any errors from reading a file not existing.

### DIFF
--- a/.local/bin/statusbar/battery
+++ b/.local/bin/statusbar/battery
@@ -20,7 +20,8 @@ esac
 for battery in /sys/class/power_supply/BAT?
 do
 	# Get its remaining capacity and charge status.
-	capacity=$(cat "$battery"/capacity) || break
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Why don't you just `2>/dev/null` the cat command instead of putting an extra test in?

Limero:
I thought that would hide the output even when successful, but you are right, this works. I have fixed it now :)

LukeSmithxyz:
Good. Yeah `>/dev/null/` is implicitly `1>/dev/null`. `1` is for standard output and `2` standard error.

PR #733, opened 2020-06-29
Add backlight control as a scroll function

Sadoon-AlBader:
This adds backlight control of monitor through scrolling.
Requires "light" pacakage and that the user can run "sudo light" without a password, easily done through sudoers edit.
Enjoy less eye strain :)

### DIFF
--- a/.local/bin/statusbar/battery
+++ b/.local/bin/statusbar/battery
@@ -10,6 +10,8 @@ case $BLOCK_BUTTON in
 🔌: charging
 ⚡: charged
 ❗: battery very low!" ;;
+	4) sudo light -A 10 ;;
+	5) sudo light -U 10 ;;
 	6) "$TERMINAL" -e "$EDITOR" "$0" ;;
 esac
 


drake-newell:
Maybe change it to use xbacklight instead? That's the LARBS default.

Sadoon-AlBader:
> Maybe change it to use xbacklight instead? That's the LARBS default.

So I tried xbacklight and light on two different laptops (Acer Aspire S7 392, HP Envy x360 13 Ryzen 2500U)
xbacklight works on the Acer, but not the HP, while light works on both.
xbacklight seems to have trouble detecting the screen on the HP.

Trebled:
Xbacklight doesn't work w/ my Acer either, thanks. 
(P.S. I found sudo to be unnecessary so long as you've appended yourself to Group video.😺)

Sadoon-AlBader:
> Xbacklight doesn't work w/ my Acer either, thanks.
> (P.S. I found sudo to be unnecessary so long as you've appended yourself to Group video.)

You're most welcome :)
Actually I just noticed today that sudo isn't necessary, but good catch on why it isn't!

PR #737, opened 2020-06-30
New Suntime module, web-api BTFO

narukeh:
Prints out Sunrise and Sunset time, and additional info in the notification box.


narukeh:
Remember to change the default url.
go to: https://www.timeanddate.com/sun/italy/rome and choose your city.
No, i cant do it like in `weather`. For countries like italy (state/city) thats ok though, but for places like USA, germany, it gives a numeric code.

PR #738, opened 2020-07-01
improve speed format

jamazi:
Convert network speed into human-readable format using numfmt, this will handle low and high values properly.


jamazi:
Can you give example about what you mean ?
I use this method in my tint2 and it works very well.

vladdoster:
polybar/polybar/issues/2127

jamazi:
It can be formatted like printf, but I personally like to keep my tint2 executor space compact as I can.

jamazi:
Actually no need to do padding in numfmt, just change :
`printf "🔻%sB 🔺%sB\\n"`
to :
`printf "🔻%4sB 🔺%4sB\\n"`

PR #753, opened 2020-07-09
Emoji Unicode 13.0 update

narukeh:
Add new stuff, not the bloat tho

### DIFF
--- a/.local/share/larbs/emoji
+++ b/.local/share/larbs/emoji
@@ -1,3127 +1,1593 @@
-😀 grinning face; U+1F600
-😃 grinning face with big eyes; U+1F603
-😄 grinning face with smiling eyes; U+1F604
-😁 beaming face with smiling eyes; U+1F601
-😆 grinning squinting face; U+1F606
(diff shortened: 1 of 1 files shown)


narukeh:
The file is `/usr/share/unicode/emoji/emoji-test.txt` from arch pkg `unicode-emoji`
I removed the bloat : 
```
sed -i '/‍/d ; /; fully-qualified/!d ; /🏻/d ; /🏼/d ; /🏽/d ; /🏾/d ; /🏿/d ; /🦰/d ; /🦱/d ; /🦳/d ; /🦲/d'
```
and then formated it with vim recordings

narukeh:
> ```
> sed -i '/‍/d ; /; fully-qualified/!d ; /🏻/d ; /🏼/d ; /🏽/d ; /🏾/d ; /🏿/d ; /🦰/d ; /🦱/d ; /🦳/d ; /🦲/d'
> ```

the ``` sed '//d'``` is actually ```sed '/<200d>/d'``` tho it will only render in vim, it concatenates emojis

PR #758, opened 2020-07-13
Update .zprofile

jamazi:
Remove trailing slash from local bin directories in PATH

This is not important, but with the trailing slash you get it doubled in some commands like:
```
type weather
weather is /home/jamazi/.local/bin//statusbar/weather
```

### DIFF
--- a/.zprofile
+++ b/.zprofile
@@ -6,7 +6,7 @@
 # to clean up.
 
 # Adds `~/.local/bin` to $PATH
-export PATH="$PATH:$(du "$HOME/.local/bin/" | cut -f2 | paste -sd ':')"
+export PATH="$PATH:$(du "$HOME/.local/bin" | cut -f2 | paste -sd ':')"
 
 # Default programs:
 export EDITOR="nvim"


remolueoend:
Hey everyone, not sure if this is the right place to comment after the PR has already been merged, but after applying this change, the subdirectories of `$HOME/.local/bin` are not added to `$PATH` anymore. Looks like calling `du` with or without trailing slash generates different outputs, at least on my machine:

image

Scripts such as `ddspawn` etc. are referenced in config files like `.config/i3/config`, where they cannot be found anymore.
Is this a general issue or did I just mess up my setup? Thanks a lot in advance for having a look.

jannec22:
Maybe you have an empty file called bin in .local/?

vladdoster:
Try doing this

```
export XDG_USER_BINARIES="${XDG_USER_LOCAL:-$HOME/.local}/bin"  # Linux

# -- add `~/.local/bin` to $PATH -- #
export PATH="$PATH:$(du "${XDG_USER_BINARIES}" | cut -f2 | paste -sd ':')"
```

jamazi:
@remolueoend your `du` looks aliased, it should not print size in human readable format by default, try rerun the command with pre-backslashed:
`\du $HOME/.local/bin`

remolueoend:
@vladdoster thanks a lot for your advice and sorry again for posting here. Next time I open a proper issue.
"Cluttered" indeed, I am still in some kind of transition from my homebrewed risotto to Luke's rice, but progress is slow..

Some answer by e-mail brought me on the right track: I use symlinks to link files from the voidrice repo into the correct directories instead of hosting the repo directly under the home folder. (Not sure how you guys are doing it, but I had a lot of trouble to keep the rice repo clean by git-ignoring everything).

Therefore `$HOME/.local/bin` is not a directory, but a symlink, which leads to a different output when calling `du` on it. I probably keep the trailing slash in my fork for now, but I need to think about how to better organize my files in the future. 

Thanks again to everyone of you, I was really surprised by the fast and supportive responses.

vladdoster:
@remolueoend you can use the `-L`/`--dereference` flag and it will fix your issue. 

Here is the `du` [man page](https://linux.die.net/man/1/du#:~:text=du(1)%20-%20Linux%20man%20page&text=Summarize%20disk%20usage%20of%20each,mandatory%20for%20short%20options%20too.&text=Display%20values%20are%20in%20units,BLOCK_SIZE%20and%20BLOCKSIZE%20environment%20variables).

PR #763, opened 2020-07-15
Fix youtube-dl parameter error

V1leDev:
This change will fix an error I've encountered when using qndl. The
download of videos would fail immediately and it seems that you have
to assign a separate variable to the parameters of youtube-dl.


LukeSmithxyz:
Eh... actually I think it's because of f820ccf4. I'll revert it.

PR #768, opened 2020-07-20
sort and add src & dt

narukeh:
I sorted them by category, so it isn't confusing, and also added `~/.local/src` & `~/.local/share`, and changed `~` for `$HOME`.

### DIFF
--- a/.config/directories
+++ b/.config/directories
@@ -1,13 +1,13 @@
 # You can add comments to these files with #
-h   $HOME
-mn  /mnt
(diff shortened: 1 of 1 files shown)


narukeh:
@vladdoster  ild, for me it looks more disordered that way. Anyhow i sorted it alphabetically since that what u guys like.

vladdoster:
imo its better to have a convention that to follow what looks good for different people. I also know whether to just use `shift + g` or stay at top of a config when I am going to edit it.

PR #774, opened 2020-07-25
Revert changes to dmenurecord

vladdoster:
Seems that @drakenewell will need to provide more details about his issue.

Someone commented that it broke for them which somewhat makes sense because alsa is lower level and might as well use it.


vladdoster:
I do apologize about this like.  I should have done more though testing. 

Seems like dmenurecord changes broke things for users. Will follow up with Drake about exact issue.

PR #779, opened 2020-07-31
Use `pkill` instead of `killall`

anntnzrb:
The `killall` command doesn't come shipped in all distributions (Void
Linux as an example), on the other hand `pkill` does.

Unsure what the exact intention on each call is but the `-9` signal can be
replaced with `-x` if the intention is not to end all instances.

Also note that I left `pkill -x Xorg` without the `-9` signal because
for some reason it won't function as expected and end up halting the
system (at least on Void Linux).

### DIFF
--- a/.config/aliasrc
+++ b/.config/aliasrc
@@ -22,7 +22,7 @@ alias \
 
 # These common commands are just too long! Abbreviate them.
 alias \
-	ka="pkill -9" \
+	ka="pkill -15" \
 	g="git" \
 	trem="transmission-remote" \
 	YT="youtube-viewer" \
--- a/.config/aliasrc
+++ b/.config/aliasrc
@@ -22,7 +22,7 @@ alias \
 
 # These common commands are just too long! Abbreviate them.
 alias \
-	ka="killall" \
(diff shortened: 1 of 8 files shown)


narukeh:
how about 
```
kill -9 $(pidof "$1")
```
😄 funny if pidof isnt shipped either 😄

anntnzrb:
> how about
> 
> ```
> kill -9 $(pidof "$1")
> ```
> 
>  funny if pidof isnt shipped either

Interestingly enough, I'm doing some tests with `pidof` as for some reason `screenkey` does not get recognized by it, but does effectively get recognized with `pgrep` instead...

anntnzrb:
Updated; the correct way of terminating is using `SIGTERM`;
`SIGKILL` should be avoided.

Source: [In what order should I send signals to gracefully shutdown processes?](https://stackoverflow.com/a/37692419)

LukeSmithxyz:
Well, even using SIGTERM, I'm not sure. It *definitely* needs `-x`, though. `pkill` will kill any process with the given match in its name without `-x`.

PR #784, opened 2020-08-01
Add termporary wm support to xinitrc

MYDavoodeh:
With this patch you can start a desktop without tweaking your dotfiles like this: `xinit ~/.xinitrc wm`, which helps with rapid development across multiple desktops.

### DIFF
--- a/.config/xinitrc
+++ b/.config/xinitrc
@@ -8,14 +8,32 @@
 # screen) and so they are needed there. To prevent doubling up commands, I
 # source them here with the line below.
 
-# export STATUSBAR="i3blocks" # Uncomment this line when using i3.
-
 . ~/.xprofile
 
 # Your default LARBS WM is determined in your `~/.profile` on login.  Here we
 # run the proper command to run when the graphical environment starts.
+# In order to temporarily overwrite the default WM, you can run something
+# like `xinit ~/.xinitrc bspwm`
 
-# Comment/uncomment these lines depending on what wm you want to use. Also
-# check the i3blocks line above.
-ssh-agent dwm
-# ssh-agent i3
+
+case ${1:-${WM:-dwm}} in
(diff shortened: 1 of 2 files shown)


.config/xinitrc
```diff
@@ -8,14 +8,32 @@
...
 # run the proper command to run when the graphical environment starts.
+# In order to temporarily overwrite the default WM, you can run something
+# like `xinit ~/.xinitrc bspwm`
+
 
-# Comment/uncomment these lines depending on what wm you want to use. Also
-# check the i3blocks line above.
-ssh-agent dwm
-# ssh-agent i3
+case ${1:-${WM:-dwm}} in
```
vladdoster, line 19: Have you tested all of these?

.zprofile
```diff
@@ -130,7 +130,7 @@ ex=🎯:\
 
 [ ! -f ${XDG_CONFIG_HOME:-$HOME/.config}/shortcutrc ] && shortcuts >/dev/null 2>&1 &
 
-if pacman -Qs libxft-bgra >/dev/null 2>&1; then
+if [ "$WM" != "dwm" ] || pacman -Qs libxft-bgra >/dev/null 2>&1 ;then
```
vladdoster, line 133: Why exclude dwm?

MYDavoodeh: Bcoz other WMs won't require libxft check.

.config/xinitrc
MYDavoodeh: Yeap I did.

.zprofile
vladdoster: I'm pretty sure you need libxft-bgra for st font rendering, it isn't a window manager issue?

MYDavoodeh: Yeap I've totally forgotten that... thanks for mentioning.

PR #796, opened 2020-08-20
add vpn indicator to internet status modul

Astronach:
for me /sys/class/net/tun*/operstate show unknown in every state, but ./net/tun* only exists, if vpn is on


MYDavoodeh:
You could also grep for vpn in /etc/resolv.conf since some of them (e.g: protonvpn) uses the DNS mechanism to address that.

PR #797, opened 2020-08-22
Add screenlock to suspend/hibernate

decantr:
Run slock and pass the command as the argument, which means the system will be
 locked on resume.

`-i` is required for this to work, which works fine on my system however I have
 `NOPASSWD: ALL` in sudoers.

### DIFF
--- a/.local/bin/sysact
+++ b/.local/bin/sysact
@@ -12,7 +12,7 @@ cmds="\
 🔒 lock		slock
 🚪 leave dwm	kill -TERM $(pidof -s dwm)
 ♻ renew dwm	kill -HUP $(pidof -s dwm)
-🐻 hibernate	${hib:-sudo -A systemctl suspend-then-hibernate}
+🐻 hibernate	slock ${hib:-systemctl suspend-then-hibernate -i}
 🔃 reboot	${reb:-sudo -A reboot}
 🖥 shutdown	${shut:-sudo -A shutdown -h now}"
 


MYDavoodeh:
Great idea!
I've been using your changes in my config files and I noticed that it won't work with anything but `slock` (and didn't even work for me with slock :|) (idk maybe I just did something wrong and this code is fine)

Eventually, I came up with this version and I thought maybe I should share it here coz it just seems more stable to me and you can run ANYTHING not only `slock` before hibernating:
```
# Add a semicolon    -\
#                     v
🐻 hibernate	slock ; ${hib:-systemctl suspend-then-hibernate -i}
```
and at the end change the \` to eval to get it working properly.
```
eval "echo "$cmds" | grep "^$choice	" | cut -d '	' -f2-"
```

Pardon me for my bad English

decantr:
> I've been using your changes in my config files and I noticed that it won't
> work with anything but `slock` (and didn't even work for me with slock :|)
> (idk maybe I just did something wrong and this code is fine)

Looking at it again, I can only imagine it's due to the removal of `sudo -A`,
 this must be causing the priveledge check to silently fail in the background.

Could you see if adding that back in, so it lookes like the below, fixes it for
 you?
```
slock ${hib:-sudo -A systemctl suspend-then-hibernate -i}
```

> Eventually, I came up with this version and I thought maybe I should share it
> here coz it just seems more stable to me and you can run ANYTHING not only
> `slock` before hibernating:

Looks good, however if I'm not mistaken I don't think there is a lock program
 env variable so it would need to modified for each program anyway?

Out of interest which lock program are you using?

MYDavoodeh:
That's the point. In my fork "lock" is an env program which is set to slock as well.

and I have checked with the sudo, also my sudoers requires no password so I don't think if it's regarding to that issue.

decantr:
Apologies for the slow reply to this.

Looking again at what you suggested, are you sure that it's working, from my
 (limited) understanding that will cause the hibernate command to be run only
 after `(s)lock` returns.

Also, you mentioned that it simply wasn't working, can you elaborate on that?

MYDavoodeh:
I realized what's the problem...
It was probably an extra quotation mark or something 😄 that I miss typed
I checked up on your original code again and it works fine... My bad! I'm sorry for the extra comments!

And yes, you are right it will run slock and then it will run hibernate but it has one main advantage: There is no need to remove sudo.

decantr:
Glad you got it working :)

> And yes, you are right it will run slock and then it will run hibernate

That's the issue though, the code you posted above will only suspend & hibernate
 _after_ you unlock your screen, which defeats the purpose.

The code in the PR will lock *and* suspespend, so you enter your password on
 resume.

> but it has one main advantage: There is no need to remove sudo.

I don't actually think we need to remove sudo, as the snippet I posted above
 *should* work, though I'm not sure how it would ask for your password from the
 lockscreen.

It's been a while since I installed LARBS, but it's possible the command is in
 the NOPASSWD list added by LARBS anyway.

adamac2:
An alternative way to do this for people using runit would be to have `zzz` run slock on resume. You would need to...

```sh
sudo mkdir -p /etc/zzz.d/resume
sudo ln -s /usr/bin/slock /etc/zzz.d/resume
```

Since `zzz` reads and executes the programs in the resume directory in alphanumeric order on resume.

decantr:
The more I hear about runit the more I like it...

I suppose adding that symlink would be best done during the LARBS setup, and
 a check here that determines the environment?

PR #801, opened 2020-08-26
Add customizability feature to icons via .zprofile

MYDavoodeh:
By changing `CF_ICONS` variable you can edit icons all across your system.

Without this you must go to more than 100 different files and edit them by hand and make sure that you update related icons (USB for example) in every file. Even by greping it takes hours (been there!)

For your changes to apply you must restart your computer or you must source your profile (`. ~/.zprofile`) in `ico` file.

For example if you want to use `material-design-icons-git` (AUR) all you
need to do is to change the variable like below:

```
export CF_ICONS="/home=󱂵:/mnt=󰡰:android=󰀲:bat-charging=󰂄:\
    bat-discharging=󰂌:bat-f=󱈑:bat-unknown=󰂃:bat=󰁹:\
    batt=󰭟:btc=󰠓:cb=📋:clock-00=󱑖:clock-01=󱑋:clock-02=󱑌:\
    clock-03=󱑍:clock-04=󱑎:clock-05=󱑏:\
    clock-06=󱑐:clock-07=󱑑:clock-08=󱑒:clock-09=󱑓:clock-10=󱑔:\
    clock-11=󱑕:clock-12=󱑖:cpu=󰻠:crypto=󰆬:disk=󰑹:\
    down=󰜮:eth=󰡪:f-move=󰪹:help=󰘥:\
    hib=󰤁:internet=󰖈:lan-x=󰌙:lan=󰌘:lbc=󰺃:leave=󰠜:lock=󰌾:\
    mail=󰇮:mem=󰍛:music=󰎇:news=󰎕:pause=󰏤:pc=󰟀:\
    pkg-down=󰏔:pkg=󰏗:q-add=󱊘:reboot=󱄋:\
    rec-cam=󰤙:rec-screen=󰊓:rec-voice=󰗋:renew=󰑓:sheets=󱈆:\
    shutdown=󰐥:sync=󰃨:temp=󰏈:tick-clock=󰾨:tick-cloud=󰅠:\
    time=󰃰:tor=󱎖:torrent-add=󰶡:torrent-seed-x=󱏽:torrent-idle=󰹢:\
    torrent-seed=󰹦:up=󰜷:usb=󰕓:vol-0=󰕿:vol-1=󰖀:\
    vol-2=󰕾:vol-m=󰕾:vol-x=󰖁:vol=󰕾:vpn=󰌆:weather-h=󰖙:\
    weather-l=󰜗:weather-perc=󰖗:weather=󰔎:\
    wifi-x=󰤮:wifi=󰤨:🌑=󰽤:🌒=󰽧:🌓=󰽡:🌔=󰽨:🌕=󰽢:🌖=󰽦:🌗=󰽣:🌘=󰽥:"
```

This model also helps with simplification of some status scripts like `disk` and `clock` (check them out now).

This is a very large commit and consumed a lot of my time.
It's natural if it's not perfect or if I missed something in the process.
I would appreciate it if you test it and check it out... And maybe let me know if I missed an icon or something.

image

These emojis are untouched (as well as commented emojis in file descriptions):
```
./.config/polybar/config:223:format-prefix = "🌡 "
./.config/polybar/config:231:format-prefix = "🧠 "
./.config/i3/config:65:bindsym $mod+Insert		exec --no-startup-id notify-send "📋 Clipboard contents:" "$(xclip -o -selection clipboard)"
./.config/lf/lfrc:49:	notify-send "🚚 File(s) moved." "File(s) moved to $dest."
./.config/lf/lfrc:60:	notify-send "📋 File(s) copied." "File(s) copies to $dest."
./.config/i3blocks/config:11:label=🎼
./.config/i3blocks/config:61:label=📅
```


lukerb52:
This looks awesome!

LukeSmithxyz:
Okay well, sure it "looks" awesome, but this is a terrible idea. I guess it's always mean when I close PRs/issues without comment, so I'll list why here because this is an illustrative case. Some reasons in order of least bad to most bad:

- Firstly, **icons are saved as environmental variables**. That means that your ability to customize any of your icons in your statusbar, for example, is as good as dead because you have to log out and log back in since they are set in your .zprofile. Moving them to the zshrc wouldn't be any better unless your calling new instances of zsh. Environmental vairables are a pain precisely because they are not easily changed. I want to get rid of them altogether.
- **All scripts have totally lost modularity.** You literally cannot download and use any of my scripts without downloading the icon script and putting it in your path and setting environmental variables (and, well restarting your WM so you can get the new variables). Every emoji will just end up being `bash: ico: command not found` on everyone's computer who wants to try my scripts.
- **This will mean massive memory and CPU usage.** It actually keeps me up at night that I still use separate shell scripts for the statusbar anyway, but think about this: instead of simply printing a character you are now calling an external script which then calls grep to search for a perl regex in an environmental variable. If you have a couple statusbar modules active and constantly updating, you might be calling grep 5-10 times *per second* **to print a character**. Mind you, there are ways of "grepping" without grep (using only shell/bash) and that hypothetically would be the right time to do it, but you're still calling an external script for this.

All of these scripts are as easy to modify as can be if you want to use font awesome or something else. You can even shift click on the statusbar modules to edit them instantly. The only theoretical benefit of such a script would be if there were hundreds of scripts all looking for the same icon choice. That isn't anywhere close to the case.

MYDavoodeh:
Thanks for your comment I appreciate it!
Yes, that's right all of the issues (especially the second case with modularity) are there. I think it was better if I listed them in the

**To anybody who has a comment on this:**
I have this merged in my repo... So, if anybody comes by with a way to work around these issues I'll be glad to discuss it and hopefully, get it to a state where its pros are more than its cons.

PR #837, opened 2020-11-08
Add support for Org-Mode files.

Epacnoss:
I know @LukeSmithxyz  doesn't really use emacs, but lots of people do, and this compiler script is just missing this one common format...
(coming from a doom-emacs user)

### DIFF
--- a/.local/bin/compiler
+++ b/.local/bin/compiler
@@ -3,7 +3,7 @@
 # This script will compile or run another finishing operation on a document. I
 # have this script run via vim.
 #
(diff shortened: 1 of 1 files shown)


Davoodeh:
Why not use `org-latex-export-latex`?
Isn't it better to have both at least? Pandoc kinda never worked for complicated documents...

Epacnoss:
> Why not use `org-latex-export-latex`?
> Isn't it better to have both at least? Pandoc kinda never worked for complicated documents...

never realised it was a thing... thanks!

PR #838, opened 2020-11-08
UAE - update autisticly efficient

narukeh:
added flags, use qndl if possible, added options, tried to make them kinda equal.
you might want to sort line 7 of `dmenuhandler` to your liking, i just sorted the short names first and kinda grouped them.

### DIFF
--- a/.local/bin/dmenuhandler
+++ b/.local/bin/dmenuhandler
@@ -2,15 +2,20 @@
 
 # Feed this script a link and it will give dmenu
 # some choice programs to use to open it.
+feed="${1:-$(printf "%s" | dmenu -p 'Paste URL or file path')}"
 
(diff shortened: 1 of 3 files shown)


AlexSyssoyev:
Hey i never used dmenuhandler, in what case it might come useful? as i see it just from an overview it looks cool for using it with xclip -o but i can't think of any other comfortable use

narukeh:
@AlexSyssoyev  you feed them links (or files), and it plays them in mpv; opens in sxiv, zathura, vim; downloads as audio or also video, with youtube-dl or curl........
i use it in newsboat, either link or dmenu; i copy a wallpaper url and `setbg` it; if its plain file, like suckless patches, i copy the link and open the content in vim immediately.

AlexSyssoyev:
Ohhh didn't thought of using it in newsboat. Great thanks! i will bind it on my config

PR #842, opened 2020-11-17
Improve `dropdowncalc` - use python with preloaded math lib

kiprasmel:
before: 

image

after:

image

### DIFF
--- a/.local/bin/i3cmds/dropdowncalc
+++ b/.local/bin/i3cmds/dropdowncalc
@@ -1,3 +1,9 @@
 #!/bin/sh
 
-ifinstalled bc && echo "Welcome to the Calculator." && bc -lq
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Actually, this script isn't used anymore since i3 was removed, so I should remove it as well.

kiprasmel:
oh no luke sorry to say but we still use i3 :/
you don't have to delete them c:

katrushenkov:
> oh no luke sorry to say but we still use i3 :/
> you don't have to delete them c:

Anyway there is no i3 in voidrice anymore.

void00r:
I think for the sake of convenience i3 shouldnt be removed. Cuz we still need straight foward easy to install and configurable WM which is not as complex as dwm and doesnt not breaks in times for crisis.

PR #851, opened 2020-11-30
fixing typo

narukeh:


narukeh:
@LukeSmithxyz  Why did you not merge? `lynx' and `Copy URL` are not functional without this patch.

PR #856, opened 2020-12-04
Added missing exit codes & minor edits

anntnzrb:
25a925f

The exit codes here should be correct (please check)

---

ff95886

I had some doubts with these ones, please check

---

245461e

---

As stated, the function already has an `exit` statement.

### DIFF
--- a/.config/sxiv/exec/key-handler
+++ b/.config/sxiv/exec/key-handler
@@ -5,12 +5,12 @@ do
         "w") setbg "$file" & ;;
         "c")
 		[ -z "$destdir" ] && destdir="$(sed "s/\s.*#.*$//;/^\s*$/d" ${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-dirs | awk '{print $2}' | dmenu -l 20 -i -p "Copy file(s) to where?" | sed "s|~|$HOME|g")"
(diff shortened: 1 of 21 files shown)


LukeSmithxyz:
Is it actually good form to use `exit 0` instead of just `exit`? I guess I don't immediately see a place here where they would be different.

Although the `exit` you removed in the `crypto` script should still be kept their because once you run the `getprices` function, it automatically signals the script to be rerun anyway. The `exit` shouldn't have an exit code there since we'd want `getprices`'s exit code to carry over.

anntnzrb:
> Is it actually good form to use `exit 0` instead of just `exit`? I guess I don't immediately see a place here where they would be different.

Well, personally I like having controll over the results of my scripts, specially if I'm using logical operators to perform certain actions based on the result of `X` command.

Not to mention I feel that using `exit` without its code is _ambiguous_, maybe like attempting to use `C`'s `exit(...)` function without a value (which is invalid), but that's me.

> Although the `exit` you removed in the `crypto` script should still be kept their because once you run the `getprices` function, it automatically signals the script to be rerun anyway. The `exit` shouldn't have an exit code there since we'd want `getprices`'s exit code to carry over.

Alright, that's why I made separate commits so you could cherry-pick; good to know.

anntnzrb:
I messed up the commits with a new addition, if you still want these changes please let know.

PR #864, opened 2020-12-13
Improve the dwmblocks clock script to update at exactly 00 seconds each minute.

frei-0xff:
After execution script schedules refresh at exactly 00 seconds.
Update interval in dwmblocks/config.h has to be 0. Update Signal assumed to be 1, as it is.
`
...
{"",	"clock",	0,	1},
...
`


narukeh:
uh nonono this is too much for a simple clock script, (i appreciate the autism though)

PR #866, opened 2020-12-15
Case block and function is more effecient

kawaiiamber:
# About
Instead of having sed check for every state, I used a case statement. Also, I abstracted the format aspect of it to its own function. When doing something like `time( for s in $(seq 10000); do; battery; done )` for this, it seems to be way more effecient with the case statement.


LukeSmithxyz:
You're reading both the status file and the capacity file twice via cat per battery per run.

kawaiiamber:
Sorry about that. I originally just kinda modified your work (big fan) and forgot about that. I commited a change and it no longer cats the info twice.

PR #879, opened 2021-01-10
Restoring feature for zsh file shortcuts

karioki:
This feature was introduced in #639.

It might have forgotten during cleaning or refactoring.


LukeSmithxyz:
Thanks. Just an oversight I guess.

PR #880, opened 2021-01-13
Use find instead du for more efficient execution.

ZaxonXP:
Find is faster and it is only one command. The last `:` is removed by the parameter expansion (works in both Bash and Zsh).


ZaxonXP:
I hope this helps. :)

otakubeam:
This seems like a minor optimization. The one which breaks compatibility too. On small systems with busybox (like alpine) find doesn't have `-printf` flag.
Can we revert this maybe?

PR #881, opened 2021-01-13
Leave a single newline if reducing trailing in init.vim

equwal:
While removing trailing newlines is a good idea, it is a problem when
editing C files which "must" have an empty line at the bottom.

So we leave just a single newline, if there were any.


equwal:
edit: tested a .c and another file.

Works by deleting all the lines then adding just one to C programs.

http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
5.1.1.2 #2

    A source file that is not empty shall end in a new-line character,which
    shall not be immediately preceded by a backslash character before
    anysuchsplicing takes place.

PR #886, opened 2021-01-17
Add keybord toggle option which needs a shortcut in dwm config

mhdzli:
[Here](https://github.com/LukeSmithxyz/dwm/pull/150)  is the `shortcut in dwm` pull request.
You can use the setxkbmap to set the default layouts  `setxkbmap -layout us,ir -option "grp:shifts_toggle,caps:escape_shifted_capslock,altwin:menu_win"` which I'm using right now in [my `.xprofile`](https://github.com/mhdzli/dotfiles/blob/master/src/.xprofile#L13),  but then you can't get the current layout using `setxkbmap`.
There is some tools that can get the current layout in this case: like `xkb-switch` and `xkb-state` .
I use [this `C` code](https://gist.github.com/mhdzli/bd017f698354ace4b788e5db8d0b7f7c) to get the current layout. You can get the compiled file from [here](https://github.com/mhdzli/dotfiles/blob/master/src/.local/bin/statusbar/sb-xkbl) or add this to `dwmblocks` as  xkb module.
And the `remaps` script also needs some changes to match the setxkpmap default options as in [here](https://github.com/mhdzli/dotfiles/blob/master/src/.local/bin/remaps#L5)

### DIFF
--- a/.local/bin/statusbar/sb-kbselect
+++ b/.local/bin/statusbar/sb-kbselect
@@ -1,11 +1,28 @@
 #!/bin/sh
 # works on any init system
 # requirements: dmenu, xorg-setxkbmap
+# Using "," as the separator set your default layouts in $HOME/.local/share/xkb-layouts
+kbselect() {\
+	kb_choice="$(awk '/! layout/{flag=1; next} /! variant/{flag=0} flag {print $2, "- " $1}' /usr/share/X11/xkb/rules/base.lst | dmenu -l 15)"
+	kb="$(echo "$kb_choice" | awk '{print $3}')"
+	}
+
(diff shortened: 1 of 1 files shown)


Davoodeh:
Interesting implementation.
In my opinion, it'd be better to avoid `grp` and other options for one be able to bind the script itself to any shortcut in dwm config or somewhere and don't worry about a second place to change a bind.
In that way, only one layout will be set on each script run. Therefore, it is possible to use `setxkbmap` to get the current layout.

The only real issue that this approach solves that my proposed implementation #635 can't do something about is that non-US layouts will not trigger keyboard events on any software that listens to KeySyms rather than KeyCodes.

`remaps` can be called on each run or removed at all in that solution. Again, no need to change that as well.

PR #887, opened 2021-01-23
Encrypted connection to Luke's RSS feed

ykahveci:
There is no reason for the URL of Luke's main RSS feed to be transferred over unencrypted HTTP. That's why i changed it to HTTPS. Nothing big, it just mildly infuriated me.


LukeSmithxyz:
I mean like every good website, mine automatically redirects to https, but fine.

PR #894, opened 2021-02-09
Implementing gpg-agent alongside ssh-agent upon starting dwm

jeanpierredevilliers:
Start a daemonized gpg-agent as a child process of the ssh-agent at login.  
The former, in turn, then starts dwm.  This is intended as an alternative to 
installing pam-gnupg, using tools already installed with LARBS.  For more 
information, refer to the relevant comments in xinitrc.

### DIFF
--- a/.config/x11/xinitrc
+++ b/.config/x11/xinitrc
@@ -14,4 +14,14 @@ else
 	. "$HOME/.xprofile"
 fi
 
-ssh-agent dwm
+# Flags that daemonize the gpg-agent as well as some 'reasonable' configuration for said daemon.
+# GPG keys are cached by the daemon for 2 hours but this duration is reset whenever the key is
+# used again by any means, this includes if it was used by a command in your crontab.
+# This is especially useful if you have mutt-wizrd schedule email synchronization via mailsync.
+# After 16 hours, however, a cached key is removed and the password will need to be re-entered.
+# The last flag is to prevent any third party caching of keys, especially on disk.
+# If you are paranoid about X-sniffing attacks, consider adding the '--grab' flag as (by
+# default_ the '--no-grab' flag is set.
+gpg_flags='--daemon --default-cache-ttl 7200 --max-cache-ttl 57600 --no-allow-external-cache'
+
+ssh-agent gpg-agent $gpg_flags dwm


BachoSeven:
@jeanpierredevilliers Would one have to explicitly set those flags even if they were already present in `~/.gnupg/gpg-agent.conf`?

jeanpierredevilliers:
> @jeanpierredevilliers Would one have to explicitly set those flags even if they were already present in `~/.gnupg/gpg-agent.conf`?

No, the configuration file is sourced by gpg-agent on startup.  Thus you are free to choose whichever you prefer.

I think your solution is more elegant in fact.  Thanks for pointing it out!

BachoSeven:
@jeanpierredevilliers I just tried using what you suggest(`ssh-agent gpg-agent dwm`, and I get this errorc(Xorg doesn't start):
```
libGL error: failed to create drawablelibGL error: failed to create drawableXIO: fatal IO error 62 (Timer expired) on X server
```

jeanpierredevilliers:
> @jeanpierredevilliers I just tried using what you suggest(`ssh-agent gpg-agent dwm`, and I get this errorc(Xorg doesn't start):
> 
> ```
> libGL error: failed to create drawablelibGL error: failed to create drawableXIO: fatal IO error 62 (Timer expired) on X server
> ```

Could you post your `gpg-agent.conf` file and perhaps your Xorg logs as well?  The latter you will find in `/var/log`.

BachoSeven:
gpg-agent configuration:
```
pinentry-program /usr/bin/pinentry-gnome3
# 1 day of cache
default-cache-ttl 60480000
max-cache-ttl 60480000
no-allow-external-cache
```
Xorg logs:
[Xorg.2.log](https://github.com/LukeSmithxyz/voidrice/files/6217030/Xorg.2.log)

BachoSeven:
I've also tried adding the `--daemon` option manually in `.xinitrc`, since you added that in this PR.
Here's the error log I mentioned earlier(i.e. the crash output from the TTY):
```
DP-1 connected 3840x2160+1920+0 (normal left inverted right x axis y axis) 600mm x 340mm
rm: cannot remove '/home/fra/.config/copyq/copyq_geometry.ini': No such file or directory
[ 28/03/2021 10:37:05.477 session_init INFO ] Switching to log file: /home/fra/.cache/picom_log
gpg-agent: a gpg-agent is already running - not starting a new one
xinit: connection to X server lost

waiting for X server to shut down X connection to :2 broken (explicit kill or server shutdown).
Gdk-Message: 10:37:05.558: copyq: Fatal IO error 2 (No such file or directory) on X server :2.

libGL error: failed to create drawable
libGL error: failed to create drawable
XIO:  fatal IO error 62 (Timer expired) on X server ":2"
      after 152 requests (152 known processed) with 0 events remaining.
(II) Server terminated successfully (0). Closing log file.
```

The first error regarding `copyq` can be safely ignored(note that I am testing this on tty2 with tty1 active with another Xorg, but I also tried it with a single Xorg session at a time, i.e. when gpg-agent is not already active).

jeanpierredevilliers:
@BachoSeven you'll want to make sure no instances of `ssh-agent` or `gpg-agent` are running before starting the X server.  This can be done with the `pgrep` command, which returns UIDs of matching processes, and if any are in fact running you can kill them with `# pkill` e.g. `# pkill gpg-agent`.  Unless you rebooted, you might also want to make sure that there isn't an X server running via `pgrep Xorg`.

You'll also definitely either want to add the `--daemon` flag tot `gpg-agent` or add `daemon` on a new line in `gpg-agent.conf`, to prevent it from running in the foreground.

If you're still stuck afterwards, it seems that X is dumping it logs to the file `~/.cache/picom_log`.  You can then post or attach it here.

BachoSeven:
@jeanpierredevilliers I've been playing around with it and I got it to work. Indeed `--daemon` was crucial
Now it's working, and the only new warnings that I've seen(i.e. from `copyq` above) are related to the `atk-bridge` which seems to be related to accessibility, and I got rid of them by setting `NO_AT_BRIDGE=1` (found [here](https://bbs.archlinux.org/viewtopic.php?id=237697)).

Btw, I'm not sure that now that I've got it working it does what the PR seemed to do, meaning I still get the `gpg` keyring prompt. Is this intended? 

the `picom` log has a bunch of `[ 28/03/2021 10:28:09.704 glx_init ERROR ] Failed to attach GLX context.` in it, the rest seems fine.

P.S. yea I used an "infinite" cache at some point because I was having issues getting `gpg-agent` to work with custom `gpg` directory location and `pam-gnupg`, so I put that insane value :'D thanks for noting

jeanpierredevilliers:
It should only prompt you for a password once per key (per reboot ofc) until the cache expires.  If it expires it will again prompt you for a password when you try to use a gpg key.

In my case I only have a single key that gets used every 10m in a cronjob that fetches my mail.  So that means the first time my mail is fetched it prompts me for my password and then the key remains cached in memory for 16h.  After 16 hours (or if I reboot ofc) I will again be prompted once for a password when my mail is fetched but this won't happen again for another 16h.

BachoSeven:
I see. I thought that with replacing `pam-gnupg` you meant the part when, at boot, the keyring unlocks using the login password(i.e. the login which allowed you to start X and dwm from xinit).
In my case, even without this change I was being prompted the first time the mail got fetched, so this doesn't really do much. Still, I learned something about how gpg-agent works, so thank you!

kronikpillow:
hm... this doesn't seem to be working for me, after i login it just pushes me back out to the login prompt

Issue #896, opened 2021-02-12
demuumount: read ignore paths from /etc/fstab


karta0807913:
Hello, I usually using another disk part mount on /var/lib/docker to storage my docker image.
When I want to umount using `demumount` script, the path `/var/lib/docker` will show up on my options.
To solve this, I suggest read the ignore path from the /etc/fstab.

There is my code:
```diff
diff --git a/.local/bin/dmenuumount b/.local/bin/dmenuumount
@@ -26,7 +26,8 @@ asktype() { \
	esac
	}

-drives=$(lsblk -nrpo "name,type,size,mountpoint" | awk '$4!~/\/boot|\/home$|SWAP/&&length($4)>1{printf "%s (%s)\n",$4,$3}')
+ignores=$(echo $(grep UUID /etc/fstab | awk '$3 !~ /swap|this/ { print $2 }') | sed 's/ /$|^/g' | sed 's/\//\\\//g')
+drives=$(lsblk -nrpo "name,type,size,mountpoint" | awk '$4!~/^'$ignores'$'$(which snap >/dev/null 2>&1 && echo "|\\/snap")'|SWAP/&&length($4)>1{printf "%s (%s)\n",$4,$3}')

if ! grep simple-mtpfs /etc/mtab; then
	[ -z "$drives" ] && echo "No drives to unmount." &&  ex
```

PR #897, opened 2021-02-12
read ignore path from /etc/fstab

karta0807913:
For different disk settings, reading the configuration path from `/etc/fstab` seems to be a good idea.
Just like me, I usually create another disk part and mount it on `/var/lib/docker` to store my docker image.
When I execute the script, the path `/var/lib/docker` will be displayed in the options.
Another problem for ubuntu users is that the `snap` daemon will install something in the `/snap` folder, and I will also ignore it.
Read settings from `/etc/fstab` can be better used in various system environments.

### DIFF
--- a/.local/bin/dmenuumount
+++ b/.local/bin/dmenuumount
@@ -26,7 +26,8 @@ asktype() { \
 	esac
 	}
 
-drives=$(lsblk -nrpo "name,type,size,mountpoint" | awk '$4!~/\/boot|\/home$|SWAP/&&length($4)>1{printf "%s (%s)\n",$4,$3}')
+ignores=$(echo $(grep UUID /etc/fstab | awk '$3 !~ /swap|this/ { print $2 }') | sed 's/ /$|^/g' | sed 's/\//\\\//g')
+drives=$(lsblk -nrpo "name,type,size,mountpoint" | awk '$4!~/^'$ignores'$'$(which snap >/dev/null 2>&1 && echo "|\\/snap")'|SWAP/&&length($4)>1{printf "%s (%s)\n",$4,$3}')
 
 if ! grep simple-mtpfs /etc/mtab; then
 	[ -z "$drives" ] && echo "No drives to unmount." &&  exit


karta0807913:
create from #896

jeanpierredevilliers:
What I like about this is that it is more *BSD friendly.  The `lsblk` command is unavailable on any of the BSDs - I believe it might be Linux only?  Hence this would also improve portability.

karta0807913:
Maybe we can read the mount point from `/proc/mounts`, then we can get rid of `lsblk`.
I write a tiny script to test it, the output just like `lsblk -nrpo "name,mountpoint"`.
I believe it works fine on usb/disk/iso mounts, but I am not sure for android mount.
Can you test the script for android mount?
If it can detect everything, we can remove the `lsblk` dependent from your script.

`little script`
```
grep -o -E '^\/[^ ]* [^ ]*' /proc/mounts
```

LukeSmithxyz:
I didn't know `lsblk` isn't on BSD. If there is an easy way to rewrite this script and it's alternate without it maintaining current features, I'd be interested.

karta0807913:
hello @LukeSmithxyz, I rewrite the entire script and remove the `lsblk`.
This script only uses the following commands:
`cat`, `realpath`, `awk`, `grep`, `sort`,`uniq`, `basename`, `tail`, `xargs`, `sudo`, and `numfmt`.
Can you please help me to check the portable and of my script?
Thank you.

usage: 
mount items:
```bash
./dmenumount
```

unmount items:
```bash
./dmenumount -u
```

Notice: I change the way to select the mount point.
You can change it at line 91-93.

```bash
#!/bin/bash

function list_partitions() {
    # list all mountable partitions.
    find /dev/disk/by-partuuid -mindepth 1 | xargs -n 1 realpath
}

function list_mounted_partitions() {
    {
        list_partitions;
        # list all mounted points
        awk '$1 ~ /^\/dev\/.*/{ print $1 }' /proc/mounts
    } | sort | uniq -d
}

function format_partition_string() {
    partitions="$(xargs -n 1 basename</dev/stdin)"
    regex=""
    for partition in ${partitions};
    do
        regex="${regex}|${partition}"
    done
    awk '$4 ~ /'"$(sed 's/\//\\\//g'<<<${regex} | cut -c 2-)"'/ { "numfmt --from-unit=K --to=si "$3|getline size; printf "/dev/%s (%s) %s\n", $4, size, $4}' /proc/partitions
}

function list_mountable_partitions() {
    partitions="$(list_partitions)";
    mounted="$({
        echo "${partitions}";
        # list all mounted points
        list_mounted_partitions
        # list all swaps
        cat /proc/swaps | tail -n +2 | awk '{ print $1 }';
    } | sort | uniq -d)";

    # partitions - mounted = mountable
    {
        echo "${partitions}"
        echo "${mounted}"
    } | sort | uniq -u | xargs -n 1 basename | format_partition_string
}

function list_mountable_android() {
    simple-mtpfs -l 2>/dev/null
}

function list_mountable_iphone() {
    if which idevice_id 2>/dev/null 1>/dev/null; then
        for uuid in $(idevice_id -l); do
            echo "${uuid} $(idevice_id ${uuid})"
        done
    fi
}

function list_mountable_items() {
    list_mountable_partitions | awk '{ print "Partition "Block $0 }'
    list_mountable_iphone | awk '{ print "IPhone "$0 }'
    list_mountable_android | awk '{ print "Android "$0 }'
}

function mount_partition() {
    partition="$1"
    mount_point="$2"
    # usually unix can choose the mount type automatically
    sudo -A mount "${partition}" "${mount_point}"
}

function mount_iphone() {
    UDID="$1"
    mount_point="$2"
    ifuse -u ${UDID} "${mount_point}"
}

function mount_android() {
    device="$1"
    mount_point="$2"
    simple-mtpfs --device "${device}" "${mount_point}"
}

function mount() {
    set -e
    items=$(list_mountable_items)
    if [ "${items}" = "" ]; then
            notify-send "Notice" "No mountable item found"
            exit 1
    fi
    choose=$(dmenu -i -p "Which item should be mount?" <<< "${items}")
    [ "${choose}" = "" ] && exit 127
    default_folder="$(echo "${choose}" | awk '{ print $NF }')"
    # I think we bouth use the systemd for the init 1 right??
    mkdir -p "${XDG_RUNTIME_DIR}/mountpoint"
    mount_point=$({ echo "${default_folder}"; find ${XDG_RUNTIME_DIR}/mountpoint/ -mindepth 1 -maxdepth 1 -type d -not -name "${default_folder}"; } | xargs -n 1 basename | dmenu -i -p "Mount at")
    [ "${mount_point}" = "" ] && mount_point="${default_folder}"
    mount_point="${XDG_RUNTIME_DIR}/mountpoint/${mount_point}"
    mkdir -p "${mount_point}"
    set +e
    read TYPE VAL <<< "${choose}"
    case "${TYPE}" in
        IPhone)
            UUID="$(cut -d ' ' -f 1 <<< "${VAL}")"
            name="$(cut -d ' ' -f 2 <<< "${VAL}")"
            notify-send "🍎 Iphone Mounting" "IPhone device ${name} mounted to ${mount_point}."
            mount_iphone "${UUID}" "${mount_point}"
            ;;
        Android)
            name="$(cut -d : -f 2 <<< "${VAL}")"
            notify-send "🤖 Android Mounting" "Android device ${name} mounted to ${mount_point}."
            mount_android "$(cut -d : -f 1 <<< "${VAL}")" "${mount_point}"
            ;;
        Partition)
            target="$(cut -d ' ' -f 1 <<< "${VAL}")"
            notify-send "💻 USB mounting" "${target} mounted to ${mount_point}."
            mount_partition "${target}" "${mount_point}"
            ;;
    esac
}

function list_unmountable_partition() {
    partion_string=""
    for partition in $(list_mounted_partitions); do
        partition_string="${partition_string}|${partition}"
    done
    {
        # list all mounted path
        awk "\$1 ~ /$(sed 's/\//\\\//g'<<<${partition_string} | cut -c 2-)/ { print \$2 }" /proc/mounts;
        # list should be ignored path (fetch from /etc/fstab)
        grep -E '^[^#]' /etc/fstab  | awk '$2 ~ /^\// { print $2 }';
    } | sort | uniq -u
}

function list_unmountable_android() {
    awk '$1 == simple-mtpfs { print $2 }' /etc/mtab
}

function list_unmountable_iphone() {
    awk '$1 == "ifuse" { print $2 }' /proc/mounts
}

function list_unmountable_items() {
    list_unmountable_partition | awk '{ print "Partition "Block $0 }'
    list_unmountable_iphone | awk '{ print "IPhone "$0 }'
    list_unmountable_android | awk '{ print "Android "$0 }'
}

function umount() {
    set -e
    items="$(list_unmountable_items)"
    if [ "${items}" = "" ]; then
            notify-send "Notice" "No unmountable item found"
            exit 1
    fi
    choose=$(dmenu -i -p "Which item should be unmount?" <<< "${items}")
    set +e
    read TYPE MOUNTPOINT <<< "${choose}"
    case "${TYPE}" in
        IPhone|Android)
            fusermount -u "${MOUNTPOINT}" && \
                notify-send "🤖 Android unmounting" "${MOUNTPOINT} unmounted."
            ;;
        Partition)
            sudo -A umount "${MOUNTPOINT}" && \
                notify-send "💻 USB unmounting" "${MOUNTPOINT} unmounted."
            ;;
    esac
}

error_message=""
if [ "$1" = "-u" ]; then
    error_message="$(umount 2>&1 > /dev/null)"
else
    error_message="$(mount 2>&1 > /dev/null)"
fi

if ! [ "${error_message}" = "" ]; then
    notify-send "Error" "${error_message}"
fi
```

PR #898, opened 2021-02-15
show red cross if wifi is disabled

futureisfoss:
show a red cross mark on statusbar if wifi is disabled/turned off.

### DIFF
--- a/.local/bin/statusbar/sb-internet
+++ b/.local/bin/statusbar/sb-internet
@@ -7,6 +7,7 @@
 case $BLOCK_BUTTON in
 	1) "$TERMINAL" -e nmtui; pkill -RTMIN+4 dwmblocks ;;
 	3) notify-send "🌐 Internet module" "\- Click to connect
+❌: wifi disabled
 📡: no wifi connection
 📶: wifi connection with quality
 ❎: no ethernet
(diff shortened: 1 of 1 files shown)


futureisfoss:
I think I should give more context for this pull request:
By default, sb-internet shows 📶 when wifi is connected to a device and 📡 when no device is connected. But there is no indication showing whether the wifi is enabled or disabled - it just shows 📡. If the wifi is enabled (turned on), it'll be constantly scanning for devices around you and its recommended to disable it (turn it off) when not in use - this saves battery, and is good for privacy/security. I think many will find it useful if there's some kind of indication on the status bar showing whether their wifi is enabled or disabled.

This PR solves this issue by showing ❌ when wifi is disabled, 📡 when wifi is enabled (but not connected to a device), and 📶 when connected to a device.

futureisfoss:
These are the commands for checking wifi information using networkmanager:
`nmcli radio wifi` - Shows whether wifi is enabled or disabled
`nmcli radio wifi on` - Enables wifi
`nmcli radio wifi off` - Disables wifi

Note: In the PR, I've used the `/sys/class/net/w*/flags` file for checking wifi status cause its more efficient and would work without networkmanager.

PR #904, opened 2021-02-25
Update sb-clock

FACKDAWURLD:
Add calendar icon to clock block so SB looks more uniform.

### DIFF
--- a/.local/bin/statusbar/sb-clock
+++ b/.local/bin/statusbar/sb-clock
@@ -26,4 +26,6 @@ case $BLOCK_BUTTON in
 	6) "$TERMINAL" -e "$EDITOR" "$0" ;;
 esac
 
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I'd rather not include the calendar emoji since it comes with a default date written on it which might cause confusion.

As a note, you could simply include the emoji in the format string, like: `date "+📅%Y %b %d (%a) $icon%I:%M%p"`.

FACKDAWURLD:
Ahh I thought the same thing. In the pull request I used the same icon used in your notifications for consistency, but for my personal config I used 🗓️ ("spiral calendar") which does not have the number in my emoji font (noto-fonts-emoji-apple). 

Nice on cleaning up the formatting too. I thought the icon would mess up the syntax if put after the "+", and it did not work if the icon went before the "+." I just updated my script with your advice. Thanks for the tip!

PR #911, opened 2021-03-10
Handle color control sequences

StephenSzwiec:
Is: `less` outputs the raw ascii of the $chartfile, which includes escape characters to change output colors. This does not resemble a graph. 
Should be: Adding -n flag to have `less` create colored output properly.

### DIFF
--- a/.local/bin/statusbar/sb-price
+++ b/.local/bin/statusbar/sb-price
@@ -21,7 +21,7 @@ updateprice() { ping -q -c 1 example.org >/dev/null 2>&1 &&
 	updateprice "$1"
 
 case $BLOCK_BUTTON in
-	1) setsid "$TERMINAL" -e less -Sf "$chartfile" ;;
+	1) setsid "$TERMINAL" -e less -Snf "$chartfile" ;;
 	2) notify-send -u low "$3 Updating..." "Updating $2 price..."
 		updateprice "$1" && notify-send "$3 Update complete." "$2 price is now
 \$$(cat "$pricefile")" ;;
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
What version of less are you using? `-n` suppresses line numbers on my version. Color displays fine as well, so I don't have this issue.

StephenSzwiec:
Strange. Version of less is 563-1 from Artix Build Bot. `less -v` gives "less 563 (PCRE regular expressions)" as the entire first line of output.

LukeSmithxyz:
By build bot, do you just mean the default in the Artix repos? That's where mine is from. While it gives the same line, on my version, it's `-V` for version.

Does `man less` say anything strange?

StephenSzwiec:
My mistake on the previous message, indeed it is `-V`. 
As for the `less` version, indeed, it is from the Artix default repos, as far as I am aware. I got the package build info using `pacman -Qi less`. 

Another mistake: my pull ought to be adding the `-r` tag. My eyes are not good. Despite this, adding `-r` fixed output for me, whereas the original `-Sf`displayed control sequences as plaintext. 

Sorry for the trouble.

LukeSmithxyz:
Oh okay. Does `-R` work for you instead?

On my system (and LARBS) I have my `$LESS` environmental variable set to
`-R` so it is always assumed on run, although if it works for you, it is
probably safe to add it to the script if used on systems without the
variable.

StephenSzwiec:
`echo $LESS` returns `-R` and adding `-R` as a flag to the script results in control character output as per the original case for me. Changing the export variable to an empty string did not change the output: I get control sequences when the line is `-Sf` and color output when I have `-Srf`

PR #912, opened 2021-03-10
update aliasrc

narukeh:
now you don't have to write `sb-` in a shell


narukeh:
btw, while were the `~sc/statusbar/` scripts renamed with a `sb-` prefix? (i didn't understand)

LukeSmithxyz:
To avoid conflicts with other programs and aliases named as generically.

PR #913, opened 2021-03-10
Add a paru support in sb-popupgrade

noisersup:
**Adds simple conditional statement which is checking if user has the paru AUR manager.**

As we all know yay is not as well maintained as in the past and many people have moved to paru. For this reason I made this commit. Of course, if the user hasn't got paru, the script will still use yay.


LukeSmithxyz:
I don't see a reason to have both. Might as well include a million other AUR helpers.

If I ever leave `yay`, I'll switch it, but I don't find that necessary now.

PR #916, opened 2021-03-20
Improve handling of pulseaudio

JaGoLi:
Hello,

I currently am using pulseaudio on a minimal setup of artix linux with dwm similar to your dotfiles.
I was always having issues with pulseaudio: it either wouldn't start correctly, would crash, or would spike at 100 percent cpu usage at startup. Reading through your closed issues it seems people are still experiencing these issues, and a solution that covers all cases has not been found.

A while back I was investigating how to start pulseaudio with brave browser and came across the simple, minimal solution that has truly fixed all my audio problems by adding 2 lines to the asoundrc file in my home directory. It basically makes pulseaudio the default audio device for alsa. Since then, I have never had any issues with pulseaudio: it has always started automatically and I don't need to make it the default audio device on a per-application basis.

I have forked the repository and already commited the changes for you to review and hopefully merge into master. Please reply to this if you have any questions.

### DIFF
--- a/.asoundrc
+++ b/.asoundrc
@@ -1,0 +1,1 @@
+.config/alsa/asoundrc
--- a/.config/alsa/asoundrc
+++ b/.config/alsa/asoundrc
@@ -1,0 +1,2 @@
+pcm.default pulse
+ctl.default pulse
--- a/.config/x11/xprofile
+++ b/.config/x11/xprofile
@@ -11,10 +11,3 @@ xcompmgr &		# xcompmgr for transparency
 dunst &			# dunst for notifications
 xset r rate 300 50 &	# Speed xrate up
 unclutter &		# Remove mouse when idle
-
-# This line autostarts an instance of Pulseaudio that does not exit on idle.
-# This is "necessary" on Artix due to a current bug between PA and
-# Chromium-based browsers where they fail to start PA and use dummy output.
-pidof -s runit &&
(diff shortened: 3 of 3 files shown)


cld4h:
Hello,
I got the the following messages after applied your commit.
And my audio is gone.
(my `$HOME=/home/hao`)

```
[hao@USB ~]$ alsamixer
ALSA lib conf.c:1229:(parse_value) default is not a string
ALSA lib conf.c:1970:(_snd_config_load_with_include) _toplevel_:2:0:Invalid argument
ALSA lib conf.c:3817:(config_file_open) /home/hao/.asoundrc may be old or corrupted: consider to remove or fix it
ALSA lib conf.c:3739:(snd_config_hooks_call) function snd_config_hook_load returned error: Invalid argument
ALSA lib conf.c:4203:(snd_config_update_r) hooks failed, removing configuration
cannot open mixer: Invalid argument
```

thatguynoe:
Try writing

```
pcm.!default {
    type pulse
}

ctl.!default {
    type pulse
}
```

to `~/.config/alsa/asoundrc` instead. (don't forget to symlink it: `ln -s ~/.config/alsa/asoundrc ~/.asoundrc`)

Perhaps the issue was incorrect syntax --- a quick [search](http://www.6by9.net/configuring-pulseaudio-for-multiple-output-devices/) found something that worked for me. Haven't had any issues so far.

PR #929, opened 2021-04-03
Change tutorialvids links to PeerTube due to potential YT deletion

abenatim:
In light of the current YouTube channel situation  (https://lukesmith.xyz/articles/deletion.html), I have changed all YouTube links from the `tutorialvids` script to use their PeerTube equivalents.

Opening PeerTube video links from `mpv` is a bit slower than YouTube or Odysee links, but I suppose that's the price one pays for not relying on third party platforms.

### DIFF
--- a/.local/bin/tutorialvids
+++ b/.local/bin/tutorialvids
@@ -5,22 +5,22 @@
 # add/remove videos from this list as I go on.
 
 vidlist="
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I'll have to test the speeds when I have better internet. I'm not opposed to them using odysee.com if it's faster for the user.

PR #930, opened 2021-04-03
improved `ifinstalled` and made `noisereduce` use it

narukeh:


narukeh:
# Explanation:

### About `ifinstalled`:
The command/binary you need, does not have the same name like the repo pkg, (e.g. opendoas provides `doas`)
You might have compiled the software yourself, therefore you have what you need to run the script, buy you cant cause you dont have the repo pkg.

----

### About `noisereduce`:
1. Made it executable
2. Made it use `ifinstalled` though, it will still output a message on the cli for those not using `notify-send`

PR #932, opened 2021-04-03
Move tutorialvids to PeerTube

ethanv2:
In light of the recent YouTube purge, I thought it would be wise to move over all the tutorial videos to their PeerTube equivalents, so that we don't get any sudden breakages if/when your channel gets hit.

It should be doing exactly the same thing, but with the PeerTube direct download links.

### DIFF
--- a/.local/bin/tutorialvids
+++ b/.local/bin/tutorialvids
@@ -5,22 +5,22 @@
 # add/remove videos from this list as I go on.
 
 vidlist="
-dwm (window manager)	https://www.youtube.com/watch?v=xnREqY-oyzM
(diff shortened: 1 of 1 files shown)


narukeh:
@ethanv2  wouldn't that be too much bandwidth for luke?
how about lbry/odysee? we know they wont remove videos, until the SEC kills them.

LukeSmithxyz:
Duplicate of https://github.com/LukeSmithxyz/voidrice/pull/929

ethanv2:
Sorry, my bad; didn't read the existing one properly

PR #935, opened 2021-04-04
little improvement for `compiler`

narukeh:
made it able to choose a preferred C compiler;
and use multiple cores/threads.

### DIFF
--- a/.local/bin/compiler
+++ b/.local/bin/compiler
@@ -13,6 +13,7 @@ file=$(readlink -f "$1")
 dir=${file%/*}
 base="${file%.*}"
 ext="${file##*.}"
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Isn't that already what `$CC` is for as an environmental variable?

narukeh:
Command cc is a symlink to gcc.
While you can `sudo make CC='clang' install`
There is nothing to do for compiling a single `.c` file into a binary.
You have to use a c compiler comand: `tcc -o gzcompat gzcompat.c`
~~THERE IS NO SUCH THING AS `cc $CC='tcc' gzcompat gzcompat.c`.~~ 
Since i already set $cc, might as well use it twice.

PR #937, opened 2021-04-05
Small improvements to power actions script

ethanv2:
I've swapped out the commands used to manage power which require root privileges for ones which either do not or can be configured as such. If SystemD is detected, the script uses `systemctl` to manage power. If not, it uses a feature of `elogind` (used on Artix) to do power actions without requiring superuser (`loginctl`).

I also added a new option to put the system to sleep, rather than hibernate. The kernel will refuse to use the sleep then hibernate option if you don't have much swap or have it disabled.

### DIFF
--- a/.local/bin/sysact
+++ b/.local/bin/sysact
@@ -5,6 +5,7 @@
 # For systemd-based systems.
 case "$(readlink -f /sbin/init)" in
 	*systemd*)
+	slp="systemctl suspend -i"
 	hib="systemctl suspend-then-hibernate -i"
 	shut="systemctl poweroff -i"
 	reb="systemctl reboot -i";;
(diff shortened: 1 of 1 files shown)


narukeh:
this is nice.

LukeSmithxyz:
Uh, does `elogind` exist on all runit/openrc machines? I don't want to use it if it will break on a machine that's not running Artix.

Check also the shutdown line. You misspelled shutdown and used `systemctl` instead.

ethanv2:
According to the [this page](https://wiki.artixlinux.org/Main/Elogind) on the Artix Wiki, all default Artix installs come with Elogind configured and setup to replace the SystemD components. So, I would assume that we don't need to worry about Elogind not being present randomly on Artix. For other distros, though - I'm not so sure.

Would you prefer it if we check for Elogind and just use root tools otherwise? It would require the entering of a password (which doesn't seem to be working correctly at the moment), but it would fix the problem for all Artix users. Thoughts?

As for the SystemD programs being used, the rationale behind that was that you can easily configure the program to not require a password. But, now I come to think about it, it is probably better to just use the normal utilities with sudo and configure it to do the same. Again, thoughts? I'm happy to change the methods if necessary.

ethanv2:
@LukeSmithxyz The more I think about this, the more this seems like the wrong way to fix the current problems with the lock screen. I'll keep having a look and try and find a better way to fix things up. In the meantime, I'm going to open a new PR for adding a sleep option to the lock screen.

PR #939, opened 2021-04-09
Fixed/improved power action script (done properly this time)

ethanv2:
I've made some small improvements in the `sysact` script, but not in a stupid way this time.

Firstly, I added the option to put the system to sleep instead of hibernate (good for people without or with low amounts of swap). Also, I added the ability for OpenRC users to sleep/hibernate the system. Finally, I fixed the annoying issue with `slock` running `sudo` as the wrong user, which should fix all the issues I was aiming to fix last time.

After [investigating](https://i.imgur.com/6ZBYnxU.png), I found that the root cause of the issues was that slock (for some reason) decides to run the commands your provide it as the "nobody" user - who isn't in the sudoers file and doesn't have SUDO_ASKPASS set. So, it messes stuff up. I fixed it by not using this feature of slock and just running the commands ourselves, with slock later.

### DIFF
--- a/.local/bin/sysact
+++ b/.local/bin/sysact
@@ -4,7 +4,7 @@
 
 # For non-systemd init systems.
 case "$(readlink -f /sbin/init)" in
-	*runit*) hib="sudo -A zzz" ;;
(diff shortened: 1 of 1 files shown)


ethanv2:
Superseeded by #1000. Closing this.

PR #945, opened 2021-05-10
dmenumount -Make empty string check more consistent

7ze:
Replace explicit empty string check condition with inbuilt  POSIX compliant empty string check flag, in adherence with the rest of the script.


narukeh:
since you are improving it, do the superlative:
```bash
test -z "$mp" && exit 1
```

7ze:
Sure. Will update it.
Is this form considered better than the `[ ]` test notation?

PR #955, opened 2021-05-29
Update: dmenumount -Make empty string check more consistent

7ze:
Replace explicit empty string check condition with inbuilt POSIX compliant empty string check flag, in adherence with the rest of the script.


7ze:
@narukeh 

> since you are improving it, do the superlative:
> 
> ```shell
> test -z "$mp" && exit 1
> ```

Updated it.

Is this generally considered better than `[ ]` notation?

narukeh:
TLDR: Yes.
Ctrl+F "Real Programmers" in  https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html
Ctrl+F "test expression" in  https://man.archlinux.org/man/dash.1
https://man.archlinux.org/man/test.1.en

7ze:
I see. Thank you for letting me know.

PR #961, opened 2021-06-17
fix sb-price: touch missing files

kiprasmel:
fixed a case where on first run, the file will not exist and the `cat` inside `printf` in line 40 will show give an error instead of doing something useful

not sure if one wants to always create the missing file, thus adjust if needed.


LukeSmithxyz:
I don't think this is a big enough issue to have the script make these checks every time the script runs. It's an issue that lasts a second.

kiprasmel:
I mean 🤷‍♀️ perhaps, it's just annoying that you have to create the files yourself

PR #964, opened 2021-06-23
fix update colorsheme on boot

johnstef99:
run wal even if there is no file as argument by looking at the true location of the bg link


krisdoodle45:
This pull request can be closed, the changes in this one are also added in https://github.com/LukeSmithxyz/voidrice/commit/9a2dc2618e4875b59a19709ec810402ce7d5df28

PR #966, opened 2021-06-30
Update volume when scrolling on volume icon.

vartiovaara:
The volume wouldn't update when scrolling on the icon without this patch. With mod++/- it updates as expected.

### DIFF
--- a/.local/bin/statusbar/sb-volume
+++ b/.local/bin/statusbar/sb-volume
@@ -4,9 +4,9 @@
 
 case $BLOCK_BUTTON in
 	1) setsid -f "$TERMINAL" -e pulsemixer ;;
-	2) pamixer -t; kill -44 $(pidof dwmblocks) ;;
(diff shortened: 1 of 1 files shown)


narukeh:
Changing the brightness in the `sb-battery` module breaks the whole bar, it "eats" the modules to the left of it.
doing this i could make it about 80% better:

```bash
	4) { light -A 5; kill -44 $(pidof dwmblocks) ;} & ;;
	5) { light -U 5; kill -44 $(pidof dwmblocks) ;} & ;;
```
doing so you put the whole command in the background, you dont wait for it to finish.

vartiovaara:
Added the "&" to this  patch. Should now work better. Thank you  @narukeh.

LukeSmithxyz:
Hmm, this shouldn't be a problem with a more recent build of my dwmblocks/dwm. The updates should happen automatically.

PR #998, opened 2021-09-03
Add partition labelling to dmenumount

rsHalford:
If a partition doesn't have a label, it results in a double space that awk ignores. Which is why the changes include a pipe into sed.


krisdoodle45:
Check the pull requests

PR #1000, opened 2021-09-05
sysact rewrite and minor stuff

narukeh:
we already have shutdown in sudo loop


narukeh:
also, congratulation to #1000 a thousand issues/pulls 🎉

narukeh:
rewrote sysact cause it was "very dumb". and made it use the `shutdown` wrapper which every init system has one, its actually already doing what was tried to do. the only diff is the sleep/hibernation way. i only know systemd and runit. someone that know others, should make a pr later on.

ethanv2:
I've closed my original PR, because this is better. I'll keep a record of how I made the script portable, though, because there is a totally portable way of doing power actions using a kernel interface. I'll do some more research on any unintended consiquences it could have.

narukeh:
@ethanv2 cool. ill checkout artix openrc when i have time. i think `loginctl` could be used for any init other than systemd, if so, it would make the script even smaller while supporting more inits.

ethanv2:
Even better than `loginctl`, we could use the /sys/power/state interface of the kernel, which allows us to manage certain system power operations really easily. It works best for suspension and hibernation.

Luke [previously said](https://github.com/LukeSmithxyz/voidrice/pull/937#issuecomment-814359512) he doesn't want to use `loginctl` if it's not present on every runit-based distro (which I don't think it nesecarily is).

Like I said, I'll keep researching on good ways to make this as portable as possible. Maybe even OpenBSD (or the *BSDs in general)?

LukeSmithxyz:
With this change, the script just `exit 1`s if you are using OpenRC or s6 or anything else. That line should be removed.

Looks like loginctl is pretty universal, so I'm not opposed to it.

Why is it a bash script? I don't immediately recognize any bashisms.

narukeh:
@LukeSmithxyz  ugh yes, fixed the shebang
i also implemented loginctl for non-systemd.

PR #1005, opened 2021-09-16
Get the location based on the ip from freegeoip

mhdzli:
If LOCATION is not set get the location based on the ip from freegeoip


mhdzli:
If `jq` is not installed on LARBS you can use `curl -sf https://freegeoip.app/xml/ | grep "City" | sed 's/<\/*City>//g'` instead

LukeSmithxyz:
I'd like to minimize, not expand, the number of unnecessary third-party background connections.

PR #1007, opened 2021-09-24
Improved passwords scripts

krisdoodle45:
Heavily based on #976, but improved.

- Used the passmenu way of getting a list of passwords instead of parsing the 'pass ls' command, which caused problems if you had passwords in folders.
- Used emojis that work by default.
- Made it more resilient to users escaping the dmenu prompt.
- Moved the view passwords dialog completely to the left, allowing you to just press the enter key to view passwords.
- Added more error messages.

In my opinion this would be a nice addition, because it also allows users to work with OTP easily.

Depends on the packages pass-otp and zbar.


cronidea:
I think this is great,

Just realised that instead of
`pass "$choice" | xclip -selection clipboard`
`notify-send "📋 Password for $choice copied!"`
`sleep 10`
`printf "" | xclip -selection clipboard`
we could just do
`pass -c "$choice"`
`notify-send "📋 Password for $choice copied!" "📋 Clipboard will be cleared in 45 seconds"`
Which is better because the clipboard functionality is built into pass

This right here is why I love open source

cronidea:
Originally I was using a script to generate a random password which used something like this
`password="$(</dev/urandom tr -dc 'A-Za-z0-9!"#$%&'\''()*+,-./:;<=>?@[\]^_{|}~' | head -c $length`
to generate a password containing a variety of characters instead of just alphanumerics

krisdoodle45:
> Originally I was using a script to generate a random password which used something like this
> `password="$(</dev/urandom tr -dc 'A-Za-z0-9!"#$%&'\''()*+,-./:;<=>?@[\]^_{|}~' | head -c $length`
> to generate a password containing a variety of characters instead of just alphanumerics

Yeah, I was confused by this as I couldn't find the script anywhere lol. I'll implement your version because that includes more characters

krisdoodle45:
That cleans the script up a lot. I've also changed the way a password is generated which might be better, but I need to find out how not to match spaces in the same command. Now I need to call truncate twice

cronidea:
I mean spaces aren't necessarily bad, I see why you wouldn't want spaces in your password but it's another character so it makes it theoretically more secure. I think the least bloated way to do this is to just leave the spaces in

Wolikjef:
> I mean spaces aren't necessarily bad, I see why you wouldn't want spaces in your password but it's another character so it makes it theoretically more secure. I think the least bloated way to do this is to just leave the spaces in

That would be the better way, but from my experience websites usually don't allow spaces in passwords sadly, I'll see what other options are

krisdoodle45:
I'll stick to your first method, that seems like the most elegant solution

cronidea:
It may be beneficial to have an "easy mode" where passwords are only alphanumeric to make them easier to type on devices with clunky UIs like TVs or even phones.

Maybe have a dmenu prompt asking and generate with `tr -dc 'A-Za-z0-9' </dev/urandom`

But the project is looking great so far!

cronidea:
I've sent you a PR with easy mode included and some POSIX-ification

krisdoodle45:
Now it also supports passwords in directories

Venem:
Looking great so far!

LukeSmithxyz:
I generally don't think it's a good idea to add graphical wrappers for command line programs that work perfectly fine as is. `pass add`, `pass rm` and `pass edit` work as efficiently as possible. And obviously `passcmd` already exists as is.

I do think there is a legitimate place for a script to scan OTP QR code secrets into `pass-otp` and to generate them on command since that often requires multiple confusing commands and copies-and-pastes (often with a graphical browser that would be resized by a terminal).

So in 58e1952ce22b364fc21763cc52e7a1418607db99, I've taken the OTP function of this script, simplified it and included it in a separate script.

Note that it assumes that all OTP are added with the `-otp` suffix (and will add this automatically). It can add OTPs, copy generated OTPs and also sync system time, since on some machines without an NTP cron, the gradual difference of system time with real time can interfere with OTPs.

PR #1010, opened 2021-09-28
Add new muteinputs script and changes to sb-volume

cronidea:
New muteinputs script can be used to mute input devices when run


krisdoodle45:
Nice idea! The muteinputs script works very well, but I've made some changes to the sb-volume script that re-adds the icons for different volume levels, and makes it more robust for people that have their microphones in a different order. 

For me the NR==2 didn't work, because awk also counts the lines that match /Built/, so I added a different counter which only prints the first microphone found that is not a monitor.

You can find the changes in the pull requests of your fork

cronidea:
Yeah the new awk command works for me too thanks

PR #1017, opened 2021-10-14
General script improvements

krisdoodle45:
- Wal: store dunst and zathura configuration files in variables for neatness.
- samedir: Just use setsid instead (or is other functionality needed?)


kronikpillow:
the samedir script doesn't work anyways @LukeSmithxyz ... it used to work with all terminals, then you changed it, and it started working only with ST, now it doesn't work with ST either

Issue #1018, opened 2021-10-16
Swich mpv hook from youtube-dl to yt-dlp


kmj9:
Youtube-dl is not active since july and youtube limits downloads to ~60 kb/s. 
The result is : watching youtube videos in mpv (newsboat youtube rss) is not posible (buffering) when quality is set above 360p.

yt-dlp solved download limits but there's no option like {  --ytdl-format=  } 

Solution would be to switch hook in mpv to yt-dlp.

krisdoodle45:
I've been having this issue myself as well, but I thought it was just youtube being odd, thanks.

I fixed it very easily, you need to install the package 'youtube-dlp' via yay and add the following to '~/.config/mpv/mpv.conf':
`script-opts=ytdl_hook-ytdl_path=yt-dlp`

It might be a drop-in for youtube-dl, in which case youtube-dl could easily be replaced system-wide, but I'll test it first.

Trebled:
Next release of mpv defaults to yt-dlp, installing alongside mpv-git in the meantime solved for me. ^_^

kmj9:
I agree. mpv-git works fine for now till mpv upgrade to the next version and yt-dlp is sensible choice.

TrevCan:
`yt-dlp` has been added to the official arch community repos.

I would recommend changing it now.

[package](https://archlinux.org/packages/community/any/yt-dlp/)

PR #1019, opened 2021-10-17
Replace youtube-dl with yt-dlp system-wide

krisdoodle45:
Relevant issue: #1018 

In essence, youtube-dl development has stopped and has caused video download speed to be capped at 60 KB/s. Replacing youtube-dl with yt-dlp which is actively maintained fixes it.


FabianPaci:
@krisdoodle45, @narukeh's  solution is definitely simpler but if you were to implement yours instead keep in mind that you missed a few commands, such as the `yta` alias for example. I think that `yta` should use `youtube-dl` instead of the predefined alias `yt` since it would avoid this issue to begin with.

PR #1027, opened 2021-10-22
fix problems with ifinstalled

futureisfoss:
simplify code & fix bug where the next argument is not checked if the first one succeeds.


futureisfoss:
https://github.com/LukeSmithxyz/voidrice/commit/0cc1758b0219704cf5a72b8e06150c54041a98e5 & https://github.com/LukeSmithxyz/voidrice/commit/2ddfa7ab0f2597b26be9872e37d7374b75a72678 solved all the issues.

PR #1032, opened 2021-11-10
-lm flag to cc to compile math.h headers

MagnusCaesar:
noticed a problem with the compiler program which didn't compile C programs with math.h headers, seemed like a problem only for math.h, this fixes the issue


thesnallygaster:
Compiling everything with math.h library linked is a stupid idea, you'd also have to include every other possible library there to solve this "issue"; that's just how C works. The best option (imo) would be to check for more arguments in C's case (as in "$#") and just include all arguments passed if possible.

PR #1036, opened 2021-11-14
Make the password prompt invisible

4G3NT:
I made the password prompt invisible by defining a foreground color, and removed the non working -P option and I also removed the echo because it's not needed, Here it is in action: https://imgur.com/a/A1eax2b


LukeSmithxyz:
If `-P` is not working, which I am about 100% sure that it is working fine, open and issue on https://github.com/lukesmithxyz/dmenu.

4G3NT:
Oh, my mistake it appears that in your fork you already have a hidden password argument I just didn't know you had a fork of dmenu 😅

Issue #1034, opened 2021-11-13
setbg script killing dunst and not replacing it


FabianPaci:
The 'setbg' script kills dunst without replacing it (e.g. setsid -f dunst). Ive fixed this with a script on my system but you might want to fix this. As far as i know, this problem only occurs when you use setbg with pywal. you can reload it other ways as well but i just killed it (killall or pkill. i used killall since pkill will terminate a program that has 'dunst' in it) and replaced it with dunst & or setsid in some cases (because dunst & doesn't work for some reason in some cases). If you need any more information just ask.

FabianPaci:
um, okay? i was just testing out a more optimal solution and reset the script and it works fine? i don't know what was wrong before. I'm sure this will happen again since i don't know what caused it so please add this to 'setbg'.

FabianPaci:
dunst config (~/.config/dunst/dunstrc) is getting reset to old config with 'setbg' with paywall.

krisdoodle45:
It's very likely it has something to do with this change:
https://github.com/LukeSmithxyz/voidrice/commit/9a2dc2618e4875b59a19709ec810402ce7d5df28 

I could personally not reproduce this, when setting my background using setbg and pywal dunst still works for me. Does dunst immediately stop working after you set a background?

FabianPaci:
Yes i t does immediately stop working. and that change is another problem where the dunst config file will reset with the old geometry that doesn't work anymore (issue #1029) .

krisdoodle45:
Could you check if ~/.config/wal/templates/dunstrc in this repo is the same as the one on your system? You might still have the version from before https://github.com/LukeSmithxyz/voidrice/commit/11b7fb7b45cdd9af8d3ea3d570986dc03779ba2e

FabianPaci:
Awesome! I cant believe I didn't try that haha, thank you for your help :)

PR #1039, opened 2021-11-14
setbg reloading dunst issue #1034

FabianPaci:
It's a hacky fix but I just wanted to propose a quick solution because it was getting really annoying when i wiped.


krisdoodle45:
This issue can be closed btw.

FabianPaci:
oops.

PR #1042, opened 2021-11-21
Make linkhandler more robust

krisdoodle45:
Sometimes sites don't add the filetype to the URL. When this is the case, the file gets downloaded from the website and gets checked locally what filetype it is. Afterwards it gets opened with the right application.


narukeh:
i dont like line 35. i think u should write the absolute path (using global env if needed). just "." is too unforeseeable, i think.

PR #1043, opened 2021-11-23
Initial changes

HelionSmoker:


kronikpillow:
none of this really makes any sense, what exactly here is a improvement? why remove pipewire conf? why remove lf? why remove vi bindkings (which work in completion btw, you hit tab, and then j k to move left right in the completion menu" ? why remove dwmblocks? why no need for audio / notifications? O.o a big wtf?

PR #1052, opened 2021-12-04
follow symlinks when detecting filetype in lf

futureisfoss:
fix bug in lf where symlinked files don't open in the expected program


futureisfoss:
If I open `~/.config/x11/xprofile` in lf, it opens the file using vim in the same window. But the same doesn't happen when I try to open `~/.xprofile` which is a symlink to the same file, this time it opens vim but in a separate window. This happens because lf doesn't properly detect the filetype of the symlinked file, instead it detects it as `inode/symlink`. This PR solves this issue by following the symlink using `readlink -f` command before detecting its filetype.

PR #972, opened 2021-07-14
update and add Diversity to emoji file!

wis:
More than annually (every several months) new Emojis are added to the Unicode Standard, this PR adds the new Emojis in 13.1, re-adds missing emojis that were removed in the last commit to this file and adds missing skin tones.

Here's the method I made and used to create the file, so it can be used for the next Unicode Standard releases:

1. open [https://unicode.org/emoji/charts/full-emoji-list.html](https://unicode.org/emoji/charts/full-emoji-list.html) and for skin tones [https://unicode.org/emoji/charts/full-emoji-modifiers.html](https://unicode.org/emoji/charts/full-emoji-modifiers.html)
2. open devtools
3. run this in Console to create and copy to clipboard part of the new file:
``` .js
copy([...document.querySelectorAll('body > div.main > table > tbody > tr:nth-child(n+4)')].map(row => {
    let code = row.querySelector('.code')?.textContent.replaceAll('U+','').trim();
    let chars = row.querySelector('.chars')?.textContent.trim();
    let name = row.querySelector('.name')?.textContent.trim();
    return code ? `${chars} ${name}; ${code};` : undefined;
}).filter(row => row !== undefined).join("\n"))
```
4.  merge and save the copied texts in a new file
5. run this command on the saved file: `cat saved-file | sort | uniq | sort -k 1,1`

[...]

Issue #844, opened 2020-11-20
Tofu in the status bar


danniel3:
After the last big update, I'm getting tofu in the status bar's battery and transmission modules. I've tried testing where it's coming in by echoing and eliminating variables in the scripts, but they don't disappear. Then I looked into the emojis themselves, turns out some of them have tofu next to them. 

Just reviewing the dmenu emoji script, these are some of the ones with tofu:
- smiling face
- frowning face
- red heart
- recycling
- telephone

Anyone else having this problem? Know any fixes?

danniel3:
On further investigation, it seems some of the emojis' monospace-ness has collapsed. All the ones with tofu have this spacing problem. 

Guess it's not a voidrice problem, but as this is the only thing I recently changed, I'm curious to know if it's more than just me, and if anyone's found a solution.

![image](https://user-images.githubusercontent.com/51640138/99787460-33f8b680-2b39-11eb-9d3e-b9d7c5af215c.png)

Snoozing101:
Same issue here.

First noticed it after some xorg updates, I thought it might be related to that.

ElectricityTaras:
Same Problem

narukeh:
@danniel3  @Snoozing101  @ElectricityTaras 
would this problem literally be fixed by deleting this <fe0f> ?
then add a `space` in the `battery` script, IF it looks too concatenated.

Snoozing101:
Thanks @narukeh that fixed it for me in the status bar.

narukeh:
@danniel3  dont forget to close the issue if it is resolved
(look at this if you missed it:  https://github.com/LukeSmithxyz/voidrice/issues/844#issuecomment-735734194 )

PR #1056, opened 2021-12-16
Update sysact

wickedjargon:
this change removes the incorrect white space char next to `shutdown` and `renew` and replaces it with a regular white space char.

The change is not fully reflected in the diff that github shows. I think github identifies both white space chars as the same char.

You're probably thinking "WTF is this guy talking about". If so, please replicate the issue by changing the default font / dmenu's font to Liberation Mono and hit Mod+BackSpace. You should see box chars next to  `shutdown` and `renew`.

### DIFF
--- a/.local/bin/sysact
+++ b/.local/bin/sysact
@@ -5,14 +5,14 @@ case "$(readlink -f /sbin/init)" in
 	*) ctl='loginctl' ;;
 esac
 
-case "$(printf "🔒 lock\n🚪 leave dwm\n♻️ renew dwm\n🐻 hibernate\n💤 sleep\n🔃 reboot\n🖥️shutdown\n📺 display off" | dmenu -i -p 'Action: ')" in
+case "$(printf "🔒 lock\n🚪 leave dwm\n♻ renew dwm\n🐻 hibernate\n💤 sleep\n🔃 reboot\n🖥 shutdown\n📺 display off" | dmenu -i -p 'Action: ')" in
 	'🔒 lock') slock ;;
 	'🚪 leave dwm') kill -TERM "$(pgrep -u "$USER" "\bdwm$")" ;;
-	'♻️ renew dwm') kill -HUP "$(pgrep -u "$USER" "\bdwm$")" ;;
+	'♻ renew dwm') kill -HUP "$(pgrep -u "$USER" "\bdwm$")" ;;
 	'🐻 hibernate') slock $ctl hibernate ;;
 	'💤 sleep') slock $ctl suspend ;;
 	'🔃 reboot') $ctl reboot ;;
-	'🖥️shutdown') $ctl poweroff ;;
+	'🖥 shutdown') $ctl poweroff ;;
 	'📺 display off') xset dpms force off ;;
 	*) exit 1 ;;
 esac


kronikpillow:
this is more of a issue to fontconfig and font settings ... I'v had that white space bar next to the sysact shutdown, also on many emoji ... but the issue is gone since i installed other fonts like 

ttf-anonymous-pro
ttf-bitstream-vera
ttf-croscore
ttf-dejavu
ttf-droid
ttf-fira-code
ttf-fira-mono
ttf-fira-sans
ttf-font-awesome
ttf-hack
ttf-inconsolata
ttf-iosevka-nerd
ttf-jetbrains-mono
ttf-joypixels
ttf-liberation
ttf-linux-libertine
ttf-mononoki
ttf-nerd-fonts-symbols
ttf-roboto
ttf-roboto-mono
ttf-ubuntu-font-family

...

wickedjargon:
I believe my commit would fix the issue for all fonts.

narukeh:
TLDR: My opinion is that this PR should be rejected.
@wickedjargon  have a read:
https://github.com/LukeSmithxyz/voidrice/pull/753
and my comments in https://github.com/LukeSmithxyz/voidrice/pull/972
https://github.com/LukeSmithxyz/voidrice/issues/844#issuecomment-735734194
https://codepoints.net/U+FE0F

Some emojis have an additional "glyph" (which do not render), so they cam modify/manipulate whatever...
e.g.:the [`<200d>`](https://emojipedia.org/zero-width-joiner/) combines two emojis. Heart on fire = heart + fire.

**long read ahead!**
replication: (this is a recurring theme) also happening e.g. with the single eye emoji
using nvim, write two lines, so you can see the difference, on the first line write "eyeEYEEMOJIeye" with no spaces: `eye👁️eye`
and on the second line, write "eye EYEEMOJI eye" with two spaces `eye 👁️ eye`
then save and `cat` the file to the terminal. You will see that the second line only has one space before the emoji, and that in the first line the second word eye gets written above the emoji.
innvim
inst
So, yeah i have put a space between the PC-emji and _shutddown_, but because how the emoji behaves, it "eats" the next character, so if you reeeally care how it looks, then yeah you should two spaces. (PC-emojiSPACESPACEshutdown)

wickedjargon:
With your explanation, it is clear you understand the issue better. I'm ok if this PR is rejected.

That being said, I just ran default LARBS (default dotfiles and progs file) and I still get the - boxes myself *with the default font and other settings*. Perhaps the issue is specific to my computer only.

narukeh:
@wickedjargon No i get the same result, not just you (with defaults and heavily modified system). What i meant is that, its just **looks**, but just make sure you actually did it right, since it can not be seen here in github 😄. Just wanted to explain what is happening.

When i wrote the emoji file there was no way of removing the `<fe0f>` tag. (but i did manage to remove the other ones)

Using just echo in the terminal i think explains it better, be sure to remove that `<fe0f>` tag
info for screenshot: first two lines dont have any spaces, last two lines have two spaces but only one gets shown. Lines 3 and 4 dont seem to have any difference hmmmmm
pic-selected-211219-2123-20

🤷🤷🤷 whatever 😄😄

kronikpillow:
like i said ... install all the above fonts i posted ... issue gone, emojis render regularly

narukeh:
@kronikpillow well the point was to know what was going on. But i did install `fft-hack`, though nothing happened, same result. And installing **every** font you listed is not a solution. The solution (i think) is to replace `<fe0f>` with two spaces, so it looks "nice", like i have explained.

wickedjargon:
Would you like to submit a commit for it?

kronikpillow:
saying this for the 3rd tiem :) i installed all the above fonts 1 by 1, i'm not entirely sure which one solved it, maybe you can try going 1 by 1 with the listed fonts, update font-cache with fc-cache -fv after every installation, refresh dwm, and see what happens till it magically fixes, then we can find out what actually caused the fix ... I am to lazy to uninstall every single font i ever installed simply to debug something that you can do yourselves by installing fonts 1 by 1 ... 

this worked for me, with every single reinstall of LARBS 

or don't ...

wickedjargon:
And what if the user decides to delete that font? I think we should choose emojis that render properly regardless of which fonts are installed. 

I would back @narukeh solution, but please submit the commit yourself.

kronikpillow:
submit a commit to install fonts? 🤣🤣🤣 jeez, forget i said anything, I was just offering you a guaranteed working solution, your thing if you use it or not 👍🏻

wickedjargon:
Yes, the prog.csv can be updated. That would be a commit

kronikpillow:
dude ... I know what a commit is ... i contributed to voidrice multiple times, why does anyone have to create a commit, for you to install fonts?

wickedjargon:
I am not using voidrice. My purpose was to improve LARBS, not seek help for my personal config.

kronikpillow:
Ok ... so I have uninstalled all my  fonts ... then I'v went 1 by 1 reinstalling them, and figured out that it is ttf-dejavu that fixes the issue ... once a user installs ttf-dejavu, the white space is  gone both from emoji script and sysact script ... so ... it's something related to ttf-dejavu in /etc/fonts/conf.d but I am unable to figure out what ... if your able to investigate this, plz report back so that we can figure it out why dejavu fixes this problem ... if your interested to find out that is ...

narukeh:
@kronikpillow hmmm that's interesting. I have always used dejavu as my fallback font, just recently i switched to noto, (__but__ i always have dejavu installed). I always had the "bug" white space, when i was using dejavu, and now that im using noto.
This is how my fontconfig loos like:
`cat .config/fontconfig/fonts.conf`
```xml
<?xml version='1.0'?>
<!DOCTYPE fontconfig SYSTEM 'fonts.dtd'>
<fontconfig>
        <alias>
                <family>serif</family>
                <prefer>
                        <family>Noto Serif</family>
                        <family>Noto Color Emoji</family>
                </prefer>
        </alias>
        <alias>
                <family>sans-serif</family>
                <prefer>
                        <family>Lexend Deca</family>
                        <family>Noto Color Emoji</family>
                </prefer>
        </alias>
        <alias>
                <family>sans</family>
                <prefer>
                        <family>Lexend Deca</family>
                        <family>Noto Color Emoji</family>
                </prefer>
        </alias>
        <alias>
                <family>monospace</family>
                <prefer>
                        <family>Terminus</family>
                        <family>Noto Sans Mono</family>
                </prefer>
        </alias>
</fontconfig>
```
At least now it got interesting, why its working for you but not for me, i would like to solve that mystery 😄😄
I'll tinker a bit, with different variations and let u know if anything happens.

narukeh:
Well as for my experience it is not ttf-dejavu. Like i said i always have that one installed. I even tried this: (also log out&in again)
`cat ~cf/fontconfig/fonts.conf`
```xml
<?xml version='1.0'?>
<!DOCTYPE fontconfig SYSTEM 'fonts.dtd'>
<fontconfig>
        <alias>
                <family>serif</family>
                <prefer>
                        <family>DejaVu Serif</family>
                        <family>JoyPixels</family>
                </prefer>
        </alias>
        <alias>
                <family>sans-serif</family>
                <prefer>
                        <family>DejaVu Sans</family>
                        <family>JoyPixels</family>
                </prefer>
        </alias>
        <alias>
                <family>sans</family>
                <prefer>
                        <family>DejaVu Sans</family>
                        <family>JoyPixels</family>
                </prefer>
        </alias>
        <alias>
                <family>monospace</family>
                <prefer>
                        <family>DejaVu Sans Mono</family>
                        <family>JoyPixels</family>
                </prefer>
        </alias>
</fontconfig>
```
still the same result as before.

Issue #1040, opened 2021-11-15
Deleting trailing whitespace doesnt set the cursor back.


FabianPaci:
Ive been trying to fix this annoying thing that happens with Luke's 'deletes all trailing white space' in his init.vim but nothing seems to work? 
![image](https://user-images.githubusercontent.com/58243358/141765112-47979df3-0678-4d6f-817f-dacee047dd92.png)
Ive trying keeping the cursors location and setting it back after its done deleting but it doesn't work. Ive been making some progress and will hopefully make a pr tomorrow morning.

narukeh:
@FabianPaci  yeah, i removed those lines just not to lose the location of the cursor. Very annoying. If you have a good solution, please post it.

FabianPaci:
ive came up with a few but they are slow. Ill keep trying.

FabianPaci:
Well that was quick. pr #1057

PR #1057, opened 2021-12-19
Resets cursor position on file save (Issue #1040)

FabianPaci:
After saving a file with some trailing white space, your cursor will be set at the line with the trailing white space instead of your initial one.


FabianPaci:
Right now it only saves the row (line) and not the column. I'm trying to fix it and ill hopefully do it soon, so don't merge it yet.

FabianPaci:
Okay, fixed!

Issue #1041, opened 2021-11-19
renew dwm, or leave dwm and relogin, breaks audio


kronikpillow:
ever since switching to pipewire to follow up on the changes to voidrice, I'v encountered problems with pipewire, 1 problem is when i try to leave dmw or refresh dwm, and relogin back in, sound doesn't work anymore until I reboot ...

2nd problem, seems that pipewire-pulse.conf is missing, hence everything related to pulseaudio is broken, including the volume icon in the statusbar ... can we get your pipewire-pulse.conf? as your pipewire settings explicitly implicate that pipewire should use that conf ... pactl and other features do not work without it

krisdoodle45:
I can reproduce the first problem

BlakePerkins:
Moving pipewire out of the autostart string in .config/x11/xprofile and instead starting it as a background task `pipewire &` resolved your first issue for me. 

Not sure about your second problem as I cannot reproduce either. What output do you get with `pacman -Qs | grep wire`

PR #1058, opened 2021-12-20
update pipewire-media-session to wireplumber

BlakePerkins:
Wireplumber is the preferred session manager for pipewire (https://gitlab.freedesktop.org/pipewire/media-session#pipewire-media-session).

### DIFF
--- a/.config/pipewire/pipewire.conf
+++ b/.config/pipewire/pipewire.conf
@@ -236,7 +236,7 @@ context.exec = [
     # but it is better to start it as a systemd service.
     # Run the session manager with -h for options.
     #
-    { path = "/usr/bin/pipewire-media-session" args = "" }
+    { path = "/usr/bin/wireplumber" args = "" }
     #
     # You can optionally start the pulseaudio-server here as well
     # but it is better to start it as a systemd service.


kronikpillow:
yes I agree :) also, Luke's approach to launching pipewire is ... flonky at best, setting up runit user service scripts is much better, and something he should consider :) I already have it setup, gonna do a PR soon

LukeSmithxyz:
I'll look into it. I have the user start Pipewire because that's what Pulseaudio more or less needed.

kronikpillow:
I'll do a git repo with my pipewire services and send you a PR, so that you can test it out ... i find this advantageous cuz of it's simple ability to restart the service when needed and similar things ... like a real service, instead of a user process ... pulseaudio works as well under it

BlakePerkins:
In the meantime, I've found
`ka pipewire && pipewire & disown`
to be a satisfactory way to restart pipewire (related to #1041).

Hooregi:
yup, this works flawlessly using `systemd` units. I've been running `wireplumber` for a while now.

eylles:
welp, i'd say runit user services and systemd user units are wrong tho, if you think about it the way desktop environments handle their daemons (gpg, ssh-agent, keyring, pulseadio, etc...) is they are spawned by the session process (same one that spawns the window manager) as they should only exist when a graphical session exists so luke's approach is rather correct.

LukeSmithxyz:
Yeah as @eylles said, I've done it this way because this traditionally how Pulseaudio *should* be run for security reasons, etc. I'm not sure if Pipewire is significantly different in this respect.

I'll make a decision on this really soon, but I am edging to keep it how it is because:

1. If it ain't broke, don't fix it.
2. It is a sheer nightmare setting up system services in LARBS since I maintain it for multiple init systems.
3. There can be a lot of permission issues with stuff like this.
4. It is a sheer nightmare setting up user/group permissions for services, especially those still in development.

Actually, as I write this out, I think it's best to keep it the way it is for now.

All patriotic Linux users should be ready to killall their audio systems. Using soystemd is kind of an increased annoyance for other reasons and really the grass is greener on the other side.

PR #1062, opened 2022-01-03
otp script: Printf is not needed here

krisdoodle45:


FabianPaci:
whats `prinf` anyway

krisdoodle45:
Yeah, it's a spelling error

LukeSmithxyz:
I want to say there was some weird use case that required this, but if there was, I've forgotten it.

PR #1063, opened 2022-01-12
Replace paru with yay in error message

nonetrix:
Since paru is not installed anymore this should say yay but instead it says paru still


nonetrix:
Umm oops seems my personal fork somehow got added here?! Why is Github so confusing to use

PR #1069, opened 2022-01-29
Move to data folder and need absolute path here

karioki:
Vimwiki is not config but it's data so ".local/share/nvim" is
    more proper location for it.

    Here relative path was given so it tries to create new vimwiki
    folder when launched from other than home dir or changing current
    dir using `:cd <path>` inside vim when launching vimwiki.


FabianPaci:
s

PR #1072, opened 2022-02-01
Shortcuts in vim commands from bm-dirs/files

karioki:
This will allow using the mappings in the vim command line.
    Here the leader is ";"

    So here `:e ;cfz` typed fast will expand into
                  `:e /home/user/.config/zsh/.zshrc`.

    This is more helpful with :sp, :vs, :cd or anywhere where a file or
    a directory is expected in the vim command line.


karioki:
A typo in commit message.  Correcting it.

PR #1075, opened 2022-02-07
Keep trailing whitespaces for *.md files

karioki:
They are syntax for the md files so they should stay.

    Also added some documentation.


karioki:
Anything we can do here?

PR #1076, opened 2022-02-17
Use the new 🪫 emoji in sb-battery

krisdoodle45:
When the battery goes below 50%, the icon now changes to this one

### DIFF
--- a/.local/bin/statusbar/sb-battery
+++ b/.local/bin/statusbar/sb-battery
@@ -9,6 +9,7 @@ case $BLOCK_BUTTON in
 ♻: stagnant charge
 🔌: charging
 ⚡: charged
+🪫: less than 50% remaining
 ❗: battery very low!
 - Scroll to change adjust xbacklight." ;;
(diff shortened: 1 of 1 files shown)


narukeh:
That seems to be a new Unicode version 14 emoji. I think only `noto-fonts-emoji` has it. When i change the font in the dmenu `config.h` to `Noto Color Emoji` it displays it.

Since i saw this, I updated the emoji file #1080 since there is a new unicode version 🙃

narukeh (commented):
You might do it cleaner like this:
```bash
test $status = '🔋' && \
test $capacity -le 50 && stat='🪫' && \
test $capacity -le 25 && warn='❗'
```
(did a pr in your fork, so it gets here)

But i think we should just replace the ❗ with the 🪫, it looks cleaner with just one emoji....

krisdoodle45:
Thanks for the improvements. In my opinion using two icons make it stand out more, which alerts the user better that the computer is low on power.

PR #1107, opened 2022-04-10
Added chmod function to `lf`

mamizaxyz:
Added a chmod function to lfrc so that you can change file modes in `lf` without the need of running another shell.
Bound it to 'Ctrl-m'


LukeSmithxyz:
This just isn't a common enough command to warrant a shortcut. Let alone one requiring user input

PR #1110, opened 2022-04-22
use setxkbmap to set menu button to super key

mhdzli:
There is an option in [`setxkbmap` man page](https://man.archlinux.org/man/xkeyboard-config.7.en) to set menu button to super (win) key.


i-c-u-p:
duplicate of #1109

PR #1118, opened 2022-05-19
use `|` as sed delimiter if pattern has `/`

i-c-u-p:
removes the need for `/` to be escaped with a backslash


LukeSmithxyz:
If you're going or replace all the `/`, you might as well use a more irregular character like `|` or something, since I could definitely imagine more urls with `'` than that.

i-c-u-p:
> If you're going or replace all the `/`, you might as well use a more irregular character like `|` or something, since I could definitely imagine more urls with `,` than that.

changed

PR #1119, opened 2022-05-23
shortcuts: make `ls` command configurable

kiprasmel:
when merging from upstream (you) it's always a hassle
to resolve the changes in this file,
but it shouldn't be, because the only modification i make
is to change the way the `ls` command behaves.

thus by extracting the SHORTCUTS_LS_CMD command out,
even if it's edited in the file,
will no longer interfere w/ other logic of the program.

fwiw, i set mine to

```sh
ls -l --all --size --group-directories-first --human-readable --classify --color=auto
```

### DIFF
--- a/.local/bin/shortcuts
+++ b/.local/bin/shortcuts
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+[ -z "$SHORTCUTS_LS_CMD" ] && SHORTCUTS_LS_CMD="ls -a"
+
 bmdirs="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-dirs"
 bmfiles="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-files"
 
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Making this exception would justify a million others.

You should alias `ls` to your extra content elsewhere (or perhaps at the beginning of your file) if you want this effect.

kiprasmel:
> Making this exception would justify a million others.

meh, i don't see why / where

> You should alias `ls` to your extra content elsewhere (or perhaps at the beginning of your file) if you want this effect.

this would work, but i don't want to change the way ls works everywhere - only for shortcuts here.
so global aliasing is not an option;
local aliasing could work, but then it does an unwanted side-effect - when i run `shortcuts`, in the current terminal session my ls command will be changed, which again i don't want.

i think the above changes are the correct place to do this.
heck, if you don't want to allow configuring the `SHORTCUTS_LS_CMD`, just always set it in the script, so that it's not coming externally (i.e. no modifications / slippery slope for justifying a million other modifications) - it still improves my situation w/o your concerns

```patch
From e4814904b51e63401cc90f7ea13c1ba052f9890b Mon Sep 17 00:00:00 2001
From: Kipras Melnikovas <kipras@kipras.org>
Date: Mon, 23 May 2022 04:01:19 +0300
Subject: [PATCH] disallow configuring via ENV variable - allow only in the
 script itself

https://github.com/LukeSmithxyz/voidrice/pull/1119#issuecomment-1134041135

Signed-off-by: Kipras Melnikovas <kipras@kipras.org>
---
 .local/bin/shortcuts | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/.local/bin/shortcuts b/.local/bin/shortcuts
index fb19dec..65df710 100755
--- a/.local/bin/shortcuts
+++ b/.local/bin/shortcuts
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-[ -z "$SHORTCUTS_LS_CMD" ] && SHORTCUTS_LS_CMD="ls -a"
+SHORTCUTS_LS_CMD="ls -a"
 
 bmdirs="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-dirs"
 bmfiles="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-files"
-- 
2.36.1
```

anything else to address?

kiprasmel:
@LukeSmithxyz 

> You should alias `ls` to your extra content elsewhere (or perhaps at the beginning of your file) if you want this effect.

this would work, but i don't want to change the way ls works everywhere - only for shortcuts here.
so global aliasing is not an option;
local aliasing could work, but then it does an unwanted side-effect - when i run `shortcuts`, in the current terminal session my ls command will be changed, and i don't want this bloat.

i think the above changes are the correct place to do this.

heck, if you don't want to allow configuring the `SHORTCUTS_LS_CMD`, just always set it in the script, so that it's not coming externally (i.e. no modifications / slippery slope for justifying a million other modifications) - it still improves my situation, while addressing your concerns?

```patch
diff --git a/.local/bin/shortcuts b/.local/bin/shortcuts
index fb19dec..65df710 100755
--- a/.local/bin/shortcuts
+++ b/.local/bin/shortcuts
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-[ -z "$SHORTCUTS_LS_CMD" ] && SHORTCUTS_LS_CMD="ls -a"
+SHORTCUTS_LS_CMD="ls -a"
 
 bmdirs="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-dirs"
 bmfiles="${XDG_CONFIG_HOME:-$HOME/.config}/shell/bm-files"
```

(i pushed this; just that it's not visible because the PR is closed)

anything else to address?

PR #1122, opened 2022-05-25
Use tmpfs for speed and I/O  reduction

xdavidel:
Right now the script writes to XDG cache directory - which is probably mounted at a physical drive.
I think it's unnecessary and might harm your disk in the long run.

It is also possible to change into /dev/shm/<filename> it case your /tmp directory is not mounted with tmpfs.


LukeSmithxyz:
I'll add this, but I think it could cause issues in the theoretically unlikely scenario when you have two desktops running this script independently.

xdavidel:
That's a good point. Maybe we should add some session metadata to the filename?

PR #1123, opened 2022-06-06
Give linkhandler option to read from clipboard

windeb204:
When passing the -c option, linkhandler reads from the clipboard. It's very useful if you have it bound as a macro, as you can copy a link in a browser, and can very easily open the copied url in a better application.

### DIFF
--- a/.local/bin/linkhandler
+++ b/.local/bin/linkhandler
@@ -6,6 +6,11 @@
 # if a music file or pdf, it will download,
 # otherwise it opens link in browser.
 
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
I changed this a little. It uses the clipboard/primary selection by default, and if given and argument, uses the argument, as before.

This eliminates the need for `-c` and also lets you bind `linkhandler` for a general plumbing tool.

windeb204:
Even better, good luck with the cleaning spree

PR #1132, opened 2022-07-01
Regions for Germany

Secondarian:
This lets you choose regions for Germany and get radar gifs from the national weather service instead of Sat24 which are way higher quality. This could also be usefull for people in other countries like France with Météo-France.


LukeSmithxyz:
Gotta say you made it way more complicated that you needed to lol. See my changes.

Either way, glad to have another country's high-quality dopplers. As this gets bigger, might want to make this less shelly.

PR #1134, opened 2022-07-03
fix wireplumber

pony-montana:
As we see in the docs here: https://pipewire.pages.freedesktop.org/wireplumber/running-wireplumber-daemon.html#pipewire-conf we need to start wireplumber in replace of the old media session, not "wireplumber as server". (This is my first hecking commit ever !!!!!!!!!)


JC-Morph:
my bad

LukeSmithxyz:
Congrats on your first hecking commiterino!

PR #1136, opened 2022-07-05
flip `&&`/`||` instead of using `!`

i-c-u-p:
untested


LukeSmithxyz:
Is there some minor (performance) reason for doing this? I think `! -f` is a lot more conducive for speed-reading and similar to natural language.

i-c-u-p:
> Is there some minor (performance) reason for doing this? I think `! -f` is a lot more conducive for speed-reading and similar to natural language.

Not that I'm aware of. Each instance is "shorter", but if it hinders readability then don't merge.

PR #1141, opened 2022-07-07
killall needs full process name to work

karioki:


LukeSmithxyz:
What system are you on? `transmission-da` is the full (albeit truncated) process name and works on my Void and Artix computers.

karioki:
It's an Arch.  For me, full process name is "transmission-daemon".  And for the truncated name `killall` doesn't work but `pkill` works.

LukeSmithxyz:
Strange that that should be different between Arch and Artix. I'll replace it with a solution that works on all distros.

PR #1144, opened 2022-07-14
Adding Accent Characters/Greek Letters/... to be copied from dmenuunicode.

0xR3V:
By adding the file `zcharacters` to `~/.local/share/larbs/chars/` the characters in that file will be displayed at the end of the list when dmenuunicode is ran. And now you can copy most used characters. Examples: 
- `½×¼=⅛`
- `it's 29°C outside`
- `1XMR = 129.47€ = 18121.63¥`
- `αβγ`


LukeSmithxyz:
Some characters here might be worth adding, but definitely not accented characters or characters from other languages. Just change the keyboard language. Adding those would open the door to adding basically anything.

PR #1149, opened 2022-07-20
Add a bluetooth manager script with dmenu using blutoothctl

mhdzli:
Modification of [rofi-bluetooth script](https://github.com/nickclyde/rofi-bluetooth) to use `dmenu` instead of `rofi`


anntnzrb:
Upon simple inspection this can be easily re-written in **POSIX sh**; also avoid using `echo` whenever possible.

appeasementPolitik:
Code where there is one if and an else and one command per condition, like at line 22,
```sh
    if bluetoothctl show | grep -q "Powered: yes"; then
        return 0
    else
        return 1
    fi
```
can also be rewritten to this:
```sh
    bluetoothctl show | grep -q "Powered: yes" &&
        return 0 ||
        return 1
```
Also "Rofi" at line 298 should become "Dmenu", and the shebang at the top should be replaced with `#!/bin/sh`.

You can easily verify POSIX compliance with the `shellcheck` command.

appeasementPolitik:
Hey, I tested this script and thought it was very useful. Didn't even know this computer had bluetooth. I also rewrote the script to POSIX sh, which you can find at the pull requests of your fork.

appeasementPolitik:
Added sb-bluetooth as well

appeasementPolitik:
Added new pull request on your fork with some bug fixes

.local/bin/statusbar/sb-bluetooth
```diff
@@ -0,0 +1,32 @@
...
+# Shows names of paired and connected bluetooth devices if bluetooth is on
+
+case $BLOCK_BUTTON in
+	1) dmenubluetooth; pkill -RTMIN+2 dwmblocks ;;
+	3) notify-send " Bluetooth module" "\- Shows paired devices.
+- Click to configure bluetooth." ;;
+	6) "$TERMINAL" -e "$EDITOR" "$0" ;;
+esac
+
+if bluetoothctl show | grep -q 'Powered: yes'; then
```
iStagnant, line 12: If the bluetoothd service isn't started when opening dwm this code will prevent dwmblocks from opening since it's waiting for a response from `bluetoothctl show` which it will never get so there should be a timeout to prevent that.
```suggestion
if timeout 0.1s bluetoothctl show | grep -q 'Powered: yes'; then
```

appeasementPolitik:
@mhdzli new commit on the PR on your fork with the fix for the above problem and some other improvements, would you consider merging?

PR #1159, opened 2022-08-09
shortened sb-volume code

Rentib:


LukeSmithxyz:
Very smart solution, but if the volume is greater than 100%, which it can easily be, no icon displays as is.

Rentib:
You are right, I totally forgot that the volume can be greater than 100%. As for the arrays I didn't know that. Thanks for letting me know.

PR #1160, opened 2022-08-10
Parse tex root for more complex latex projects

ssnailed:
Allows parsing texroot comments so that you can compile a LaTeX document from outside of the document root, as long as it contains the usual magic comment format:
% !TEX root = doc.tex


LukeSmithxyz:
This will be very useful, but I've added with changes:
- Uses a separate script so this functionality can be extended to `opout`.
- Just use `% root =` for potential use in other filetypes in the future.

PR #1172, opened 2022-08-21
Enable passing of CLI flags in .xprofile

i13e:
This PR allows users to add command line flags to auto-started programs in the xprofile sh "array". A few common use-cases for this I can think of would be starting things such as `emacs --daemon` and `picom --experimental-backends`. 

I use this format in my own dotfiles and have tested it with LightDM and TTY login and it completely functions. Do note that each new program must be on a separate line and `pgrep -fl` must be used since you are querying the full command string including flags (not just the command itself). $program cannot have quotes around it either otherwise the shell interpreter takes the command literally.

For more information, see:
[ShellCheck: SC3030 - In POSIX sh, arrays are undefined](https://www.shellcheck.net/wiki/SC3030)
NOTE: I tried using the first option in this link and it simply does not work on login with either LightDM or TTY.

Please let me know if you have any feedback as this is my first PR. Thanks!


narukeh:
i think the only reason it went to a for loop was that the xprofile file would go to a few lines, since all programs would be in one variable separated with spaces. Your version seems nice, but then you wouldn't need a loop at all, just execute programs line after line, just like in the old xprofile.

PR #1181, opened 2022-08-31
add nightlight script and dmenublocks action

bubstance:
Just a little option to toggle a "nightlight" mode. Requires `xsct`.

### DIFF
--- a/.local/bin/nightlight
+++ b/.local/bin/nightlight
@@ -4,8 +4,6 @@
 # toggle for a "nightlight" color temperature
 # edit value of settemp to your desired color temperature (minimum is 700)
 
-ifinstalled "xsct" || exit
-
 settemp="2700"
 colortemp="$(xsct | cut -d' ' -f5)"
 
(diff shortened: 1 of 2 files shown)


pineapples5972:
It would be good to have some statusbar icon showing its on. Even tho its obvious its on but turning battery icon into crescent moon wont hurt. just suggesting.

bubstance:
Would adding the moon at the end be a fair compromise? It feels a bit off to completely replace the other status indicators (not charging, charging, stagnant, etc).

appeasementPolitik:
On my system I have redshift-minimal installed from the AUR autostarted from ~/.xprofile which is also an option

public-void-ryan:
Moons show up twice if you have multiple batteries
pic-selected-220903-1225-56
.

bubstance:
> Moons show up twice if you have multiple batteries.

Ah, probably should have seen that coming.

I don't have any machines with multiple batteries, so I don't really have a way to test locally. Does moving it to the end of the line do anything?

`printf "%s%s%d%%" "$status" "$warn" "$capacity" ; unset warn; echo "$night"` or something to that effect?

public-void-ryan:
> done && printf "$night" "\\n"

yes this works perfect!

PR #1197, opened 2022-09-30
Changed ~ to $HOME

Eloitor:


LalleSX:
It does the same thing. (see pic)
pic-selected-221028-1606-48
Is there a specific reason to make it $HOME instead?

PR #1207, opened 2022-10-05
Update getbib

emrakyz:
The script does not work with pdf files because most of the pdf files have a space after "DOI:" argument and that space causes curl errors. I have tried this way and it works. 

image


appeasementPolitik:
I tried replacing both commands with one command for fun, and can you try replacing

`grep -io "doi:.*" -m 1 | sed 's/ //g'`

with

`sed -n '/[dD][oO][iI]:/{s/.\{4\} \?\([^ ]*\).*/\1/p;q}'` and seeing if it also works in your case?

While I was testing I also found cases where the DOI line looked like this:

`DOI: 10.0000/S0000000000000000 · Source: Something` 

which this solution also covers.

PR #1209, opened 2022-10-07
Update Getbib Again

emrakyz:
Hi Luke! getbib script is amazing for academic work. From the last pull request of mine I understand that you don't want to use another program while we can do the same thing with just one program (just grep without sed).

The error is sometimes we get a "." at the end of the doi. This time I've done my homework by watching your video called "Regular Expressions: Regex" and done the changes accordingly :)

Output of pdftotext | grep, Before the change: `doi:10.1016/j.cell.2012.04.034.`
Output of pfttotext | grep, After the change: `doi:10.1016/j.cell.2012.04.034`


LukeSmithxyz:
Fixed in 228cc6cfdbdca5a04cc7e98289b447ab100d8133

PR #1215, opened 2022-10-14
Enabling Gruvbox Theme & True Color Support for Nvim

emrakyz:
Gruvbox better suits the system since terminal's color scheme already is Gruvbox.

The changes also enable the truecolor support which is natively supported by st. The colors look better and distinguishable this way.

The last line with autocmd is for transparency. Without it, transparency wouldn't work.

I hugely recommend trying this out. You'll maybe commit the changes later.


admbeck:
morhetz/gruvbox plugin is pretty big and can cause a bunch of bugs. There are more minimal gruvbox themes available for vim out there like vim-gruvbox8

emrakyz:
I haven't encountered with any bugs and it's the one that made the color scheme popular. Thanks for informing though.

PR #1220, opened 2022-10-28
Make the shell remember your ssh key

LalleSX:
In the video [Setting up a Website...](https://youtu.be/3dIVesHEAzc?t=555) Luke talks about not having to type the password every time you ssh into your server. 
However when you follow the tutorial it prompts the password for the ssh-key every time instead, this line removes that need to ask every time and only once.


LalleSX:
There is also a [pull request](https://github.com/LukeSmithxyz/LARBS/pull/492#issue-1427310663) to add 'keychain' to LARBS

PR #1225, opened 2022-11-02
'above' and 'below' as direction options

gh0ste:


LukeSmithxyz:
This will break the lower portion of the script since it automatically decides the tertiary display by `grep -v` the secondary display out of all of the other options.

nitsvga:
I apologize for the long time I left this undone. 

I wanted to, but I really cannot try the tertiary piece since I have no third display, but by side tests simulating the actual outcome; seems like it should work fine!.. If not I'll check it out.

Sorry for the mistakes, I'm kind of a newbie. Anyway, I cannot close this without thanking you for your channel and general contributions Luke!

PR #1232, opened 2022-11-11
Adding bookmarkthis script

warthog18:
Adding bookmarkthis script by Luke Smith in .dotfiles repository... do you agree?
Thanks!

### DIFF
--- a/.local/bin/bookmarkthis
+++ b/.local/bin/bookmarkthis
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+bookmark="$(xclip -o)"
+file="$HOME/.local/share/larbs/snippets"
+
+if grep "^$bookmark$" "$file"; then
+	notify-send "Oops." "Already bookmarked!"
+else
+	notify-send "Bookmark added!" "$bookmark is now saved to the file."
+echo "$bookmark" >> "$file"
+fi


Inlife-raphael:
Is it possible to add the DWM binding also?

warthog18:
> Is it possible to add the DWM binding also?

Just uncomment the line with the binding of MODKEY Shift and in the command add something like
{ MODKEY|ShiftMask,		XK_b,		spawn,		{ .v = (const char*[]){ "bookmarkthis", NULL } } }

warthog18:
> > Is it possible to add the DWM binding also?
> 
> Just uncomment the line with the binding of MODKEY Shift and in the command add something like { MODKEY|ShiftMask, XK_b, spawn, { .v = (const char*[]){ "bookmarkthis", NULL } } }

The file to modify is config.h under src/dwm

emrakyz:
I have improved the bookmarking function, thinking you would like it. Take a look at it, I will open a new PR.

You only get the names of the websites on dmenu. For example you only see "Searxng". If you enter, you can search things with a keyword without a browser opened. If the chosen site doesn't need you to search, then you open the site immediately.

With this way, we improve readability since we remove the unnecessary part from the urls.
Instead of seeing "https://www.paulgo.io/search?q=", we see "Searxng"

Since we need to separate the urls and the names of the websites, we use a different command "jq" for that and the formatting of the website file is a little bit different yet still simple.

PR #1234, opened 2022-11-13
Added tar.gz preview capability

warthog18:
Added tar.gz preview capability, that was missing. Only zip were supported.


LukeSmithxyz:
Since it's the same command, the expression for both tar.gz and zip can be used on one line.

PR #1235, opened 2022-11-15
Improve sd script

futureisfoss:
Now sd will also work on apps like mpv and zathura

### DIFF
--- a/.local/bin/sd
+++ b/.local/bin/sd
@@ -2,10 +2,8 @@
 
 # Open a terminal window in the same directory as the currently active window.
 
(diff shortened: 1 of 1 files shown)


aymey:
`cd: /proc/[0-9]+/fdinfo: No such file or directory` when the app isn't compatible, previously it would create a new terminal. (Example tested on `firefox`)

futureisfoss:
I thought this wouldn't be an issue since we can simply open a terminal instead of using the `sd` script on Firefox. But there's an easy fix for this, wait a minute.....

futureisfoss:
I just fixed it, now it'll open a new terminal when you run it on Firefox, no more `cd` errors.

PR #1239, opened 2022-12-12
fixed shortcut keymap conflicts in lf

0n3W4y7ick3t:
Bug details: the _shortcuts_ script generates cd and edit shortcuts for lf from bm-files/dirs, but in the original version, `C` is already mapped to function _copyto_, so when pressing `Cxx`, instead of _cd xx_, the leading `C` triggers _copyto_ prompt first. Same goes for `Exx`.
So I changed them to be `<c-c>xx` and `<c-v>xx`, by default `<c-c>` and `<c-v>` map to nothing.

### DIFF
--- a/.local/bin/shortcuts
+++ b/.local/bin/shortcuts
@@ -39,5 +39,5 @@ awk "!/^\s*#/ && !/^\s*\$/ {gsub(\"\\\s*#.*$\",\"\");
 	printf(\"abbr %s \42\$EDITOR %s\42 \n\",\$1,\$2) >> \"$fish_shortcuts\"  ;
 	printf(\"map %s :e %s<CR> \n\",\$1,\$2)          >> \"$vifm_shortcuts\"  ;
 	printf(\"map %s shell \$EDITOR %s \n\",\$1,\$2)  >> \"$ranger_shortcuts\" ;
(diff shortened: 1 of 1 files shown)


0n3W4y7ick3t:
when pressing C:
image

when pressing \<c-c\>:
image

0n3W4y7ick3t:
I changed edit commands to be starting with <c-v> instead(v for vim). By default <c-v> maps to nothing.

0n3W4y7ick3t:
I found another problem: keymaps that are prefix of others will mask others,  eg. when marked /mnt as mn, and ~/Music as m, you can't do \<c-c\>mn to cd /mnt, because the leading \<c-c\>m will cd ~/Music first.
It seems there is no way other than, like, to reset keymaps to mn and mu.

TheYellowArchitect:
I have a fresh LARBS installation. All posts above the last one are no longer true (<c-v> and <c-c> output unknown mapping)

As for the latest post, I don't understand. <c-c> and <c-v> isn't mapped anywhere in lf by default. The commited request from my side (fresh LARBS installation) has
`ranger_shortcuts="/dev/null"` and `lf_shortcuts="${XDG_CONFIG_HOME:-$HOME/.config}/lf/shortcutrc"`

PR #1240, opened 2022-12-12
make zsh complete ignore case

0n3W4y7ick3t:
`$ ls bOO<Tab> `
=>
`$ ls Books`


aymey:
Sometimes its nice if its case sensitive though. I think the best option would be to search like vim does, where irs not case sensitive unless you include a capital letter or something similar.

PR #1241, opened 2022-12-13
add .local/bin/dmenupasspy

SingularisArt:
This is a simple python script that displays all your passwords, from the `pass` program to dmenu.
Then, once you select one, it'll be in your clipboard for 5 seconds. After that, your clipboard is cleared.
I didn't want to spend hours writing it up in bash since I'd have to google shit. So, I decided to write it up in some python magic. Feel free to re-write it to bash or something.


LukeSmithxyz:
passmenu

PR #1243, opened 2022-12-28
Check if user hasn't selected anything in keyboard layout switch script

mukkumayc:
If we don't check and user presses escape in dmenu it will print nothing.


mukkumayc:
I mean it will set no layout with setxkbmap and then it won't find layout on update in dwmblocks.

PR #1249, opened 2023-01-15
Beautify mediainfo notification

2084x:
before:
pic-selected-230115-1109-40

after:
pic-selected-230115-1109-21


LukeSmithxyz:
This kind of makes it harder to pick out the variable information.

One idea might be to put it on one line as you do, but also to make the variable information bold.

PR #1251, opened 2023-01-17
Clean code for sd script

futureisfoss:


futureisfoss:
I made more changes to the code, now its easier to add more programs to the exclude variable. Having to exclude a few programs is one side effect of making sd script work on all programs like zathura, mpv, etc.

futureisfoss:
Actually, upon further testing I figured out a way to make this work without having to manually exclude programs. I will create a new PR when I'm done writing the script, for now I'm closing this PR.

PR #1255, opened 2023-01-27
Don't stop mounter script on ls error

appeasementPolitik:
If no decrypted LUKS drives are found, the script errors with `ls: cannot access '/dev/disk/by-id/dm-uuid-CRYPT-LUKS2-*': No such file or directory` and stops, this makes sure the error doesn't stop the script.

### DIFF
--- a/.local/bin/mounter
+++ b/.local/bin/mounter
@@ -16,7 +16,7 @@ alldrives="$(lsblk -rpo "uuid,name,type,size,label,mountpoint,fstype")"
 # Get all LUKS drives
 allluks="$(echo "$alldrives" | grep crypto_LUKS)"
 # Get a list of the LUKS drive UUIDs already decrypted.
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Working for me without any decrypted drives. If you're having issues are you sure this is it?

appeasementPolitik:
I did some more debugging by adding `trap 'echo "ERROR: $BASH_SOURCE:$LINENO $BASH_COMMAND" >&2' ERR` at the top of the script, which shows the command the script stops on.  It wasn't `ls`, but actually `grep` that errors for me when a non-LUKS USB drive is plugged in.

LukeSmithxyz:
Oh yeah. That makes sense. I guess I didn't have the issue because my main hard drive is always encrypted so the variable is never null.

PR #1259, opened 2023-02-04
Fix mouse issue in firefox, add avif icon and add DK & SE to dopplar

LalleSX:
###  Fix issue with right click in window managers

Mousebros...

This issue might disappear over time as more packages are installed, but i had this issue when i first installed LARBS and this fixed it.
Also mentioned in the  [arch wiki](https://wiki.archlinux.org/title/firefox#Right_mouse_button_instantly_clicks_the_first_option_in_window_managers
)

### Add Denmark and Sweden to weather radar

_Duplicate because the other was deleted due to cleaning up my github gits_

### DIFF
--- a/.config/firefox/larbs.js
+++ b/.config/firefox/larbs.js
@@ -42,3 +42,6 @@ user_pref("network.http.referer.XOriginPolicy", 0);
 
 // Disable Firefox sync and its menu entries
 user_pref("identity.fxaccounts.enabled", false);
+
+// Fix the issue where right mouse button instantly clicks
+user_pref("ui.context_menus.after_mouseup", true);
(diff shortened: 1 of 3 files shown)


LukeSmithxyz:
The Swedish selection shows Denmark and the Denmark option seems not to work... Might be an error on there side, but I don't see a reason to add them if they're dysfunctional.

LalleSX:
Wired, it worked for me when I had it on Denmark.
Welp I removed it now, so everything should be working fine.

LalleSX:
There where also all ready `EU: SCAN: Scandinavia` I just thought it would be better to have the individual countries listed
I have now listed the Scandi countries in the name section so when you search on either Denmark, Sweden and Norway it comes up.

PR #1261, opened 2023-02-06
Automate remaps upon new keyboard

ssnailed:
This PR would make remaps be run automatically when a new keyboard is plugged in. Note that if the script will be run unprivileged, you will have to add `kernel.dmesg_restrict = 0` to /etc/sysctl or to a file in /etc/sysctl.d. This allows unprivileged reading of dmesg which probably fits the threat model of the majority of desktop users.

`dmesg -W -f kern | grep "input:" -q` blocks until dmesg writes a line containing "input:", which happens when a keyboard is plugged in. I'm sure you could refine this grep statement, but I left it as general as possible.

### DIFF
--- a/.config/x11/xprofile
+++ b/.config/x11/xprofile
@@ -8,7 +8,7 @@ setbg &			# set the background with the `setbg` script
 #xrdb ${XDG_CONFIG_HOME:-$HOME/.config}/x11/xresources & xrdbpid=$!	# Uncomment to use Xresources colors/settings on startup
 remaps &		# run the remaps script, switching caps/esc and more; check it for more info
 
-autostart="mpd xcompmgr dunst unclutter pipewire"
+autostart="mpd xcompmgr dunst unclutter pipewire remapd"
 
 for program in $autostart; do
 	pidof -s "$program" || "$program" &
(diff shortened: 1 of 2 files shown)


LukeSmithxyz:
Wouldn't `dmesg` need to be run as root?

ssnailed:
Not with `kernel.dmesg_restrict = 0` set in /etc/sysctl. Some distros have this set by default.

LukeSmithxyz:
Well that's systemd-specific. Since I and probably most others using
this are running Artix, I'd like to have this enabled by default on the
different init systems before adding something like this script.

ssnailed:
Actually as far as I know this exists on artix and even some BSDs. These are just config files for the kernel.

LukeSmithxyz:
Oh okay. You're right. I had tried it in `/etc/sysctl` and it didn't work on Artix/runit, but `/etc/sysctl.conf` works fine. All add this to the LARBS repo and then add this script.

ssnailed:
My apologies, I made a typo

ssnailed:
Yeah for some reason it doesn't work without the sleep, I assume there's some lag between the kernel event and the x server picking up the new keyboard

ssnailed:
There may be a better way to accomplish this, I'll look into it tomorrow.
This post might lead me down the right path
https://superuser.com/questions/717263/how-udev-sends-input-device-change-notification-to-xorg#755533

PR #1266, opened 2023-02-11
unmounter: prompt to eject drive if possible

appeasementPolitik:
Whenever the last mount point of a drive gets unmounted, prompt to fully eject the drive as well. Mostly useful for external hard drives, as this gracefully stops them from spinning. Doesn't prompt if any partition of the device is still mounted somewhere.

It doesn't prompt to eject when unmounting an Android device, and I've also tested it to work when unmounting LUKS drives, even if some of the mounted partitions of the drive are encrypted and some not.

It might also be possible to do the ejecting passwordless, but I haven't checked yet.

### DIFF
--- a/.local/bin/unmounter
+++ b/.local/bin/unmounter
@@ -23,6 +23,17 @@ notify-send "Device unmounted." "$chosen has been unmounted."
 # Close the chosen drive if decrypted.
 cryptid="$(echo "$lsblkoutput" | grep "/${chosen#*/}$")"
 cryptid="${cryptid%% *}"
-test -b /dev/mapper/"${cryptid##*/}"
-sudo -A cryptsetup close "$cryptid"
-notify-send "🔒Device dencryption closed." "Drive is now securely locked again."
+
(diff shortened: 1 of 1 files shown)


appeasementPolitik:
I found out how to make it passwordless. It is done by adding:

`/bin/sh -c echo\ offline\ >\ /sys/block/[a-zA-Z0-9]*/device/state;\ echo\ 1\ >\ /sys/block/[a-zA-Z0-9]*/device/delete`

to `/etc/sudoers.d/01-larbs-cmds-without-password`, and replacing line 38 with:

`sudo sh -c "echo offline > /sys/block/$devlabel/device/state; echo 1 > /sys/block/$devlabel/device/delete" &&`

LukeSmithxyz:
These lines are just way too ugly to warrant a feature I don't think anyone will want anyway.

More importantly, dealing with another useless prompt you have to respond to every time you unmount a drive would annoy me. The reason I rewrote the scripts was originally to lessen the prompts the user got.

PR #1272, opened 2023-02-15
fixes #1270

### LINKED ISSUE #1270: sb-volume doesn't account for changing volume from dwmblocks
If you use the scroll wheel to change the volume from dwmblocks or middle click it to mute, no changes show in dwmblocks because sb-volume doesn't send a kill signal. Was there a reason for not adding a kill signal?

aymey:
is this stupid? we shouldn't have to manually kill?? (also this happened with `pamixer` i cant believe no one mentioned this as an issue for that long, unless its not an issue.

### DIFF
--- a/.local/bin/statusbar/sb-volume
+++ b/.local/bin/statusbar/sb-volume
@@ -4,7 +4,7 @@
 
 case $BLOCK_BUTTON in
 	1) setsid -f "$TERMINAL" -e pulsemixer ;;
-	2) wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle ;;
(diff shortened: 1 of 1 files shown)


iStagnant:
I don't think it's a dumb solution i already though of this but i asked to check if there was reason it was not done, and you should use `kill -44 $(pidof dwmblocks)` instead of `pkill` since it's more efficient and this is how the signal is invoked [dwm/config.h](https://github.com/LukeSmithxyz/dwm/blob/master/config.h#L154).

iStagnant:
I just tried to middle click the icon and use the scroll wheel and it updates even without this and for some reason it updates(?) idk why it didn't initially and why it did now.

PR #1280, opened 2023-02-22
Improve window PID discovery in sd script

ssnailed:
The current way of finding the PID of the active window is quite fragile due to the extremely loose regex in place. Since xdotool is already a dependency of multiple other scripts I'd recommend just using that.


LukeSmithxyz:
Nevermind, I misread something.

PR #1296, opened 2023-04-02
Make dwmblocks modules async

### LINKED ISSUE #1295: Backgrounding sb-forecast/sb-moonphase
Currently when dwmblocks starts, it will hang on sb-forecast/sb-moonphase if wttr.in is slow to respond. The easy solution to this is to just introduce a timeout by using the -m or --connect-timeout flags for curl. Alternatively, you could have sb-forecast immediately return but have it start a background process to fetch the weather. This sub-process would then have to send a signal to dwmblocks once the weather is fetched. This would considerably improve the start time of dwmblocks for people with chronically slow internet connections and would prevent dwmblocks from hanging on startup. 

Is this a change that would make sense? I would be happy to contribute this if this is something you are interested in.

appeasementPolitik:
fixes #1295

Makes the dwmblocks modules that make a request to the internet asynchronous so they don't block the entire status bar from showing on slow or no connection. The bar is actually quicker even when the internet is pretty fast which is nice.

For `sb-iplocate`, add 

`/* {"",	"sb-iplocate", 0,	27}, */` 

to the `config.h` of dwmblocks.


.local/bin/statusbar/sb-forecast
```diff
@@ -48,6 +48,4 @@ case $BLOCK_BUTTON in
 	6) "$TERMINAL" -e "$EDITOR" "$0" ;;
 esac
 
-checkforecast || getforecast
-
-showweather
+checkforecast && showweather || { ( getforecast && pkill -RTMIN+5 "${STATUSBAR:-dwmblocks}") & echo; }
```
ssnailed, line 51: Maybe it would make sense to use -RTMIN+${1:-5} instead of just -RTMIN+5 so that the signal number could be passed as an argument in the dwmblocks configuration

LukeSmithxyz: Yeah, that's a really good idea, especially considering otherwise you would have to standardize and hard-code the currencies used in `sb-price`.

LukeSmithxyz:
I think it'd be much better for the signal code be passed as the fourth argument to the `sb-price` module. This is for people who occasionally run it in non-status bar circumstances, so you shouldn't need to include the code.

appeasementPolitik:
Does anything else need to be changed?

appeasementPolitik:
Updated to latest commits, also added an exponential backoff retry which tries five times to connect but doubles the amount of seconds between each try starting from 1 second. This is because sometimes `sb-forecast` wasn't able to make a connection when the wifi was still starting up on boot and this fixes it.

appeasementPolitik:
I made some changes. For the backgrounding, first it's checked whether there is even internet at all, and if there is, curl is tried. If either of them fail both of these checks are retried after 60 seconds. 

Because of this change, it's now possible to start your computer up without internet, and if at some point you get internet access, the curl downloads will automatically be completed after a little while, after which the backgrounding stops. The same happens if the services are initially down and later come up again. You don't need to refresh dwm for the downloads to be retried.

I've also made sure that no new backgrounding processes are started whenever dwm is refreshed and the backgrounding is still active.

appeasementPolitik:
@LukeSmithxyz bump

josephkerkhof:
@appeasementPolitik I am trying to get these new statusbar scripts working on my machine, but can't seem to get any output in the status bar. I cannot get any output from calling the scripts from the terminal either. Do I need to do anything else to get them to work?

I have added the following to my dwmblocks `config.h` file:
```
	{"",	"sb-iplocate",				30,	27},
```

appeasementPolitik:
Strange, you shouldn't have to do anything extra as far as I'm aware. These are the relevant dwmblocks config entries for me:

```
{"",	"sb-price xmr Monero 🔒 24",	9000,	24},
{"",	"sb-price eth Ethereum 🍸 23",	9000,	23},
{"",	"sb-price btc Bitcoin 💰 21",	9000,	21},
{"",	"sb-moonphase",	18000,	17},
{"",	"sb-forecast",	18000,	5},
{"",	"sb-iplocate", 0,	27},
```

I personally have `sb-iplocate` only run at dwm (re)start by setting the time to 0, but that shouldn't make a difference.

It's normal that most scripts don't give an output and immediately return whenever the download is not completed yet. It starts a service in the background that tries the download. One thing you could check is to run the scripts and open `htop`, and look for either `curl`, or `sleep 60` services, as these get run in the background by the scripts. If these are running it would mean that curl is constantly failing, probably because the servers are down.

TheYellowArchitect:
Off-topic: @josephkerkhof I installed fresh Artix+LARBS last week, dwmblocks functioned perfectly from the get-go. This issue is unrelated to any problem you have (open an issue and I will try to help if not others)

On-topic: This is a great addition, what holds it from merge?

appeasementPolitik:
@thetubster99 I implemented the changes on your branch for fast retry, works very nicely. It now very quickly tries 100 times first when the network is still going up before switching to a cooldown of 60 seconds. For the case where the network adapter takes a while to go up, to avoid unnecessarily sleeping for 60 seconds.

appeasementPolitik:
> Off-topic: @josephkerkhof I installed fresh Artix+LARBS last week, dwmblocks functioned perfectly from the get-go. This issue is unrelated to any problem you have (open an issue and I will try to help if not others)
> 
> On-topic: This is a great addition, what holds it from merge?

I'm not sure what's holding it from merge. I've been using this PR for a year now without problems.

Issue #1260, opened 2023-02-05
sb-volume


elvekdarzhinov:
https://github.com/LukeSmithxyz/voidrice/blob/master/.local/bin/statusbar/sb-volume#L29

This line trims only one leading zero which causes the number to be interpreted as octal which doesn't work for 08, 09.

PR #1303, opened 2023-04-16
Fix sb-volume decimal integer misinterpreting

EvgenHi:
Shell script interprets zero-prefixed decimal integer as octal integer so it throws an error when "vol" is equal to "08" or  "09" because this values are invalid octal numbers. Now "vol" forced to be decimal.

### DIFF
--- a/.local/bin/statusbar/sb-volume
+++ b/.local/bin/statusbar/sb-volume
@@ -30,9 +30,9 @@ split() {
 vol="$(printf "%.0f" "$(split "$vol" ".")")"
 
 case 1 in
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
What shell are you using for `/bin/sh`? The trailing zeros are already removed in dash and bash by the split function and the printf command.

LukeSmithxyz:
Also, I should add that you are adding a bashism here.

EvgenHi:
I found the problem. The commit from 7 Feb  (#1260) fixed the issue I was trying to resolve. I just didn't looked for that. Thank you.

PR #1308, opened 2023-04-22
An Extremely Simple and Efficient Script I Think You All Like

emrakyz:
EDIT: Superseded by #1419 

The inspiration: (Luke Smith's Video): [**"Unix Chad JUST WON'T STOP Piping!"**](https://youtu.be/w_37upFE-qw) 

**Execution Time**: **0.0 seconds** on my system with more than **10.000** video files.

Very Short Summary: 
- This script feeds the "dmenu" with a list of all videos in a hard drive. 
- **File paths and extensions are removed** for better readability for the dmenu list. 
- Opens the chosen video file in "mpv" using the complete paths and extensions. 
- An approach of having the least amount of lines & characters while keeping the simplicity and performance has been used.

`The actual script is actually a 3 line script but I added some error handling to increase convenience. You can delete everything but the last 3 lines if you want.`

The first 20 lines of the script checks if related programs installed, and check if you have a video database for locate command. If not it does everything for you and notifies you appropriately. If you can't run sudo commands without entering your password, then it will also notify you about it regarding that you need to run the script on terminal or change your permissions.

### Detailed Explanation for Everyone:

**locate** command is for locating all of the video files with their given extensions:
**-d ~/.config/.mymlocate.db**: Specifies the path to the database to use.
**-b**: Matches only the name of the files (no directory names).
**-r ...** : Uses a regular expression to match filenames. Specifically, it's matching files that end with the given extensions.
**> "$tempfile"**: This redirects the output (the matched file paths) of the locate command to the temporary file we created earlier.

**chosen_name=$(sed 's|.*/||; s/\.[^.]*$//' "$tempfile" | dmenu -p "Select Video")**:
**'s|.*/||'** This sed expression (details below) removes the directory part of each file path, leaving just the base filename.
**s/\.[^.]*$//** This sed expression removes the file extension from the base filename.

**chosen_path=$(fgrep "/$chosen_name\." : The pattern to search for. It looks for lines that contain our chosen filename followed by a period, ensuring we match the correct file. fgrep is faster than grep because it doesn't have regex processor. It's also safer so you can open file names with regex characters.

Finally we use the chosen path with mpv.

### Regular Expressions

We combine two sed substitution commands into one command by separating them with **;**

We use this scheme here: **'s|pattern|replacement|; s|pattern|replacement|'**

**'s|.*/||'**

**s** Stands for "substitute".

**|** This is an alternative delimiter used for the substitution operation. Traditionally, the **/** character is used as the delimiter, but in this case, **|** has been chosen likely to avoid having to escape the **/** character in the pattern, making it more readable.

**Pattern: .*/**

**.*** This matches any character (represented by .) zero or more times (represented by *).

**/** This matches a forward slash. Since we use **|** for the separator we don't have to escape the slash.

So, **.*/** matches everything up to the last **/** in a string.

For example, for the string **/home/user/videos/movie.mp4**, the pattern **.*/** matches **/home/user/videos/**

Replacement: **||**

There's nothing between the two **|** delimiters, which means it's an empty string.

This command will replace everything up to and including the last **/** with nothing (it'll remove that part), effectively extracting just the base filename from a full path.

**'s/\.[^.]*$//'** 

We use traditional scheme here **s/pattern/replacement/**

**s** As before, this is the "substitute" command.

**/** This time, the traditional **/** delimiter is used.

**\.[^.]*$**

**\.** Matches a literal dot (.). The dot is prefixed with a backslash to escape it, since a dot has special meaning in regex (it matches any character).

**[^.]** The square brackets denote a character class. The **^** at the beginning of this class negates the character class, meaning it will match any character that is not a dot.

**\*** Matches the preceding pattern (in this case, the negated character class) zero or more times.

**$** Matches the end of the line.

This pattern matches the last dot and everything after it (up to the end of the line) that doesn't contain another dot. It's designed to match file extensions.

For example, in the filename movie.part1.mp4, the pattern will match .mp4.

**Replacement: //**

Again, there's nothing between the two **/** delimiters, so this part will be replaced with nothing.

This command will remove the last file extension from a filename.

When you run both commands in sequence using the semicolon to separate them, you're effectively:

Removing the path, leaving only the base filename.
Removing the file extension from that filename.

For the string **/home/user/videos/movie.part1.mp4**, after both **sed** operations, you'll be left with **movie.part1**

### DIFF
--- a/.local/bin/videosearch
+++ b/.local/bin/videosearch
@@ -1,6 +1,6 @@
 #!/bin/dash
 
-command -v locate >/dev/null || {
+command -v locate dash >/dev/null || {
     sudo pacman -S mlocate &&
     notify-send "Locate not found. Installing..." || {
         notify-send "Failed. Run the script once on terminal OR change sudo permissions."
@@ -11,7 +11,7 @@ command -v locate >/dev/null || {
 [ -s "$HOME/.config/.mymlocate.db" ] || {
     notify-send "You have no database. Creating it..."
     disk_path=$(echo "" | dmenu -l 0 -p "Enter the disk path (e.g '/mnt/harddisk'): ")
-    doas updatedb -o ~/.config/.mymlocate.db -U "$disk_path" || {
+    sudo updatedb -o ~/.config/.mymlocate.db -U "$disk_path" || {
         notify-send "Failed. Run the script once on terminal OR change doas permissions."
         exit 1
     }
--- a/.local/bin/videosearch
+++ b/.local/bin/videosearch
@@ -10,7 +10,7 @@ command -v locate >/dev/null || {
 
 [ -s "$HOME/.config/.mymlocate.db" ] || {
     notify-send "You have no database. Creating it..."
-    disk_path=$(echo "" | rofi -dmenu -l 0 -p "Enter the disk path (e.g '/mnt/harddisk'): ")
+    disk_path=$(echo "" | dmenu -l 0 -p "Enter the disk path (e.g '/mnt/harddisk'): ")
     doas updatedb -o ~/.config/.mymlocate.db -U "$disk_path" || {
         notify-send "Failed. Run the script once on terminal OR change doas permissions."
         exit 1
@@ -18,5 +18,5 @@ command -v locate >/dev/null || {
 }
 
 video_files=$(locate -d ~/.config/.mymlocate.db -b -r '.*\.\(mp4\|mkv\|webm\|mov\|m4v\|wmv\|flv\|avi\)$')
-chosen_file=$(echo "$video_files" | sed 's|.*/||; s/\.[^.]*$//' | rofi -dmenu -p "Select Video")
+chosen_file=$(echo "$video_files" | sed 's|.*/||; s/\.[^.]*$//' | dmenu -i -l 10 -p "Select Video")
 mpv "$(echo "$video_files" | grep -F "/$chosen_file.")"
--- a/.local/bin/videosearch
+++ b/.local/bin/videosearch
@@ -1,10 +1,4 @@
 #!/bin/sh
-
-tempfile=$(mktemp)
-locate -d ~/.config/.mymlocate.db -b -r '.*\.\(mp4\|mkv\|webm\|mov\|m4v\|wmv\|flv\|avi\)$' > "$tempfile"
-
-chosen_name=$(sed 's|.*/||; s/\.[^.]*$//' "$tempfile" | dmenu -i -l 10 -p "Select a video file:")
-chosen_path=$(grep "/$chosen_name\." "$tempfile")
-
-mpv "$chosen_path"
-rm -f "$tempfile"
+video_files=$(locate -d ~/.config/.mymlocate.db -b -r '.*\.\(mp4\|mkv\|webm\|mov\|m4v\|wmv\|flv\|avi\)$')
+chosen_file=$(echo "$video_files" | sed 's|.*/||; s/\.[^.]*$//' | dmenu -p "Select Video")
+mpv "$(echo "$video_files" | grep "/$chosen_file.")"
--- a/.local/bin/videosearch
+++ b/.local/bin/videosearch
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+temp_file=$(mktemp)
+locate -d ~/.config/.mymlocate.db -b -r '.*\.\(mp4\|mkv\|webm\|mov\|m4v\wmv\|flv\|avi\)$' | awk -F/ '{name=$NF; gsub(/\.(mp4|mkv|webm|mov|m4v|wmv|flv|avi)$/, "", name); print name "\t" $0}' > "$temp_file"
+chosen_file_name_and_path=$(cut -f1 "$temp_file" | dmenu -i -l 10 -p "Select a video file:")
+mpv "$(awk -F'\t' -v chosen="$chosen_file_name_and_path" '$1 == chosen {print $2; exit}' "$temp_file")" && rm "$temp_file"
--- a/.local/bin/videosearch
+++ b/.local/bin/videosearch
@@ -1,4 +1,22 @@
-#!/bin/sh
+#!/bin/dash
+
+command -v locate >/dev/null || {
+    sudo pacman -S mlocate &&
+    notify-send "Locate not found. Installing..." || {
+        notify-send "Failed. Run the script once on terminal OR change sudo permissions."
+        exit 1
+    }
+}
+
+[ -s "$HOME/.config/.mymlocate.db" ] || {
+    notify-send "You have no database. Creating it..."
+    disk_path=$(echo "" | rofi -dmenu -l 0 -p "Enter the disk path (e.g '/mnt/harddisk'): ")
+    doas updatedb -o ~/.config/.mymlocate.db -U "$disk_path" || {
+        notify-send "Failed. Run the script once on terminal OR change doas permissions."
+        exit 1
+    }
+}
+
 video_files=$(locate -d ~/.config/.mymlocate.db -b -r '.*\.\(mp4\|mkv\|webm\|mov\|m4v\|wmv\|flv\|avi\)$')
-chosen_file=$(echo "$video_files" | sed 's|.*/||; s/\.[^.]*$//' | dmenu -i -l 10 -p "Select Video")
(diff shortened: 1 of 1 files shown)


TheYellowArchitect:
`updatedb` command output: 
sudo: updatedb: command not found ((maybe it has something to do with lingering sudo rights after installation))

Anyway, requires the installation of `locate` to avoid the above problem.

The command `updatedb -o ~/.config/.mymlocate.db -U /mnt/externaldrive/Videos`  which contains around 1 TB was surprisingly fast, didn't even take a minute.

I can confirm it works as is, but I suggest the following 3 additions:
1. Make it case-insensitive. If some word starts with capital (youtube video names don't follow proper grammar) I can't find it
2. Have many choices show, instead of 1 covering from the end of input to the right edge of the screen (dropdown would also be preferable)
3. When you press exit, it shouldn't output:

```bash
fgrep: warning: fgrep is obsolescent; using grep -F
[file] Cannot open file '': No such file or directory
Failed to open .

Exiting... (Errors when loading file)
```

emrakyz:
You can use my other script on another pull request named "fixnames" to make your file names Unix compatible and increase your consistency. It's extremely fast.

fgrep can be changed with grep -F to remove that error but it's harmless.

This script can be made more robust in general but I prefer simplicity for jobs like this instead of extreme correctness: "[Worse is Better](https://en.wikipedia.org/wiki/Worse_is_better)". 

And thanks but no. I didn't just watch that video and write the script :). I am very interested in C and shell scripting even though I am not a programmer myself. I try to write extremely minimalist scripts by achieving the same targets. I generally prefer Dash and I use Bash only if I feel like I have to use it.

emrakyz:
Here is your example site. Arch Linux website use this format. So you will highlight the link below. You don't need to copy. You need to highlight it (I don't do it on the video because I use another command that works with copying that is `wl-copy`. x-clip also has an equivalent but in that video Luke uses `xclip -o` so I kept the same):

**https://archlinux.org/packages/?sort=&q=**

So we will use this format in our file (the script does this automatically):
Arch Packages=https://archlinux.org/packages/?sort=&q=

Let me walk you through it with a video. I will back-up all of my current files and delete everything then start from 0. I just add another keyword for this site because it doesn't have search in it. Then I add it into my bookmarks and try to search in it.

[showcase_arch.webm](https://github.com/LukeSmithxyz/voidrice/assets/89175311/f7773ac0-8e7c-4d51-baeb-59dab7a5f7a9)

emrakyz:
@TheYellowArchitect 

By the way, on Rofi, there is an option called **-auto-select** that selects the matching entry if it's the only one, automatically without enter like you mentioned for "@@". On Dmenu as far as I know, it's not possible but it can be made possible with patches. I can't do anything about it in the script. Shell scripting is technically not similar to programming. We can't change how exclusive programs behave.

By the way you can find and open filenames when they have **Uppercase letters and spaces**. They are not a problem. The problem is that we use sed (because it's lighter and more efficient than other programs in this task) with regular expressions. So when your file names have things that can be interpreted as regular expression characters then you may not find or open them. But uppercase letters or spaces won't cause a problem. We can use awk or another approach here but it's an edge case and there is no need for that. Users can do the changes themselves if needed since this script is extremely minimal and therefore extensible.

TheYellowArchitect:
>By the way, on Rofi, there is an option called -auto-select that selects the matching entry if it's the only one, automatically without enter like you mentioned for "@@". On Dmenu as far as I know, it's not possible but it can be made possible with patches. I can't do anything about it in the script. Shell scripting is technically not similar to programming. We can't change how exclusive programs behave.

I understand. I myself barely wrote for dmenu, so I didn't know if dmenu can do accomplish that.

>This script can be made more robust in general but I prefer simplicity for jobs like this instead of extreme correctness: "[Worse is Better](https://en.wikipedia.org/wiki/Worse_is_better)".

I agree that for developers or specific users, simplicity is preferred instead of adding many error checks. It's how programs bloat after all (gotta cover 0.1% edge use-case)
Honestly, it's 3 lines so keep it as is, I was personally surprised when I read the wall of text expecting some huge script, then see 3 lines. Makes anyone happy to see it, because after all, computers were meant to be simple and easy to configure.

>And thanks but no. I didn't just watch that video and write the script :). I am very interested in C and shell scripting even though I am not a programmer myself. I try to write extremely minimalist scripts by achieving the same targets. I generally prefer Dash and I use Bash only if I feel like I have to use it.

I am happy to hear it, and generally to see more and more people who aren't full-time programmers, use programming to aid their daily activities, whatever these are. Kinda like Luke
Btw with your knowledge in C and shell-scripting, I would argue you are a programmer, you just aren't paid for it lol

>EDIT: I removed the error you see on the terminal. I made the dmenu filtering case insensitive. Have fun!

Thank you very much. I have an external drive which contains ~1 TB of videos (I basically hoard all good videos, and also have one for "to-watch later"), this script is very helpful, it never passed my mind to make something like this, thanks :D

emrakyz:
Here you can find the "fixnames" script "It's a very small script again with good explanations": #1322 

I think you will like this too. These scripts can also be good for you if you try to understand shell scripting or improve your skills. There are interesting ideas in them.

For the comment "recommendation", I generally prefer robust documentation instead of adding it to the script as a comment but for this it can be added of course. I can also make it automatic but we need to make sure dmenu prompt works for you to put the second input.

TheYellowArchitect:
>Here you can find the "fixnames" script "It's a very small script again with good explanations": https://github.com/LukeSmithxyz/voidrice/pull/1322

I would use it, except `a) Remove non-English characters.` would remove greek and japanese characters, so it would cause problems for me. Seems pretty good though.

>These scripts can also be good for you if you try to understand shell scripting or improve your skills. There are interesting ideas in them.

Indeed. These days I am configuring this computer, and I had to learn the basics of bash (still not familiar with much stuff), the last 4~5 days have been great

The best part I think is when you open some random command in the same config file you are writing (or usually in .local/bin) and basically understand how it functions. This is true for all code ofc, you use what works and is often-used in the codebase, it's the fastest way to learn (trial&error and understanding why it gave error, repeat)

Related to the above is that the best documentation is the code itself, it should be so clear that it's self-explanatory. Ideally, the code would be so clean that the function names and variable names are descriptive and accurate of their functions (and good coding practices like a function having no more than 2 depths)

The above cannot be applied to bash however, as it was made in a different time, and it's goal is to be as small and performant as possible. chain operations instead of if/else, pipes instead of proper input parameters, and the infamous obfuscated variable names or esoteric symbols which have rules of their own.
For this reason, I suggest a simple comment, because I believe the user who will use this script, will browse these .local/bin scripts out of curiosity what he has available. He will stumble into this, read it, and run the command, job done. No need to auto-run for him or anything, these 3 lines are clean.

As for robust documentation, documentation which is seperated from the code is bad imo, users never bother searching for it (also such documentation decays rapidly, but luckily not here since it's just 3 lines without intention of addition)

emrakyz:
I would still recommend naming your files with English character equivalents instead of using Greek or Japanese characters. It makes things much more consistent and less problematic to deal with. I am not sure how easy it would be though for your use case. In Turkish for example, it's easy to replace "çatı" with "cati".

emrakyz:
@TheYellowArchitect 

> Btw for merging it, I suggest just adding a comment at the top of the bin, saying basically:
#run `updatedb -o ~/.config/.mymlocate.db -U /VideosPath` to populate the database to be parsed below

I automated the process. If dash and locate are not installed, the script will install them for you and give you related notifications.

If they are installed but there is no database, then it will prompt you asking in dmenu for the path of the disk with your videos. You can enter "/mnt/harddisk" here for example. Then it will continue updating the database.

So we covered the error cases here and it's much more intuitive now. But it's 22 lines instead of 3 :) The efficiency of the script won't change though. The checks are pretty minimal and most of the code won't run if everything is okay.

TheYellowArchitect:
22 lines instead of 3 spooked me, but seeing the commited file, its still basically 3 lines, seperated by 2 clear condition codeblocks, so it's a good change. Though the locate part could be just 1 comment line which asks the user, and the database part uses rofi which should be included in the pacman locate codeblock. Those 2 codeblocks could be 2 comment lines instead, but no harm is done, the main part of the script (3 lines) works the same and I would feel bad to say the additions weren't required.

emrakyz:
Yes I separated them to make it automatic and clear for its appearance and ease of change.

Yes. 22 lines are not valid for most users. The script basically only runs these checks. So if you pass these, the script is 3 lines for you :)

```
command -v locate >/dev/null
[ -s "$HOME/.config/.mymlocate.db" ]
```

emrakyz:
@TheYellowArchitect You may want to check #1419

PR #1311, opened 2023-04-23
Update volume module on headphone (dis)connect

### LINKED ISSUE #1304: sb-volume doesn't update mute state when headphones are inserted
Hello to You All,
on Thinkpad X220, Artix configurd with LARBS as of april 2023, dwm updated to 6.4.
Volume controls, mute command and scroll are perfectly working with wpctl command.

Steps to reproduce:

1) without headphones plugged, set mute ON. Icon in statusbar is correctly updated
2) plug headphones: the statusbar icon remains in mute (incorrect). Play a video and you will listen to audio (correct because the audio was muted ONLY without headphones). If you push the volume keys, volume icon is updated.
3) unplug headphones: icon in statusbar remains in mute, and the audio is muted (correct)

To summarize: the hardware is working correctly (it retains the state mute/unmute without/with headphones respectively).
The issue is the update of the icon in the statusbar....

Any clue? Thanks a lot!

appeasementPolitik:
fixes #1304 

Requires a new application in progs.csv, `acpid`. There is a problem of how to enable this daemon to start on boot from install, especially considering that on artix, this application also has its own -runit, -6, etc. packages (maybe there's a different package or way of doing this better?)

Another issue of volumed is that it depends on a hardcoded signal for dwmblocks which can't be passed as its ran from xprofile, but seeing that other scripts like transadd also use a hardcoded signal, I think it is alright.

If anyone finds any better way of listening to acpi messages, please give a comment.

### DIFF
--- a/.config/x11/xprofile
+++ b/.config/x11/xprofile
@@ -7,7 +7,7 @@ xrandr --dpi 96		# Set DPI. User may want to use a larger number for larger scre
 setbg &			# set the background with the `setbg` script
 #xrdb ${XDG_CONFIG_HOME:-$HOME/.config}/x11/xresources & xrdbpid=$!	# Uncomment to use Xresources colors/settings on startup
 
-autostart="mpd xcompmgr dunst unclutter pipewire remapd"
+autostart="mpd xcompmgr dunst unclutter pipewire remapd volumed"
 
 for program in $autostart; do
 	pidof -sx "$program" || "$program" &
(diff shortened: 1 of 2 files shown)


appeasementPolitik:
Another option would be to append `acpid` to /etc/rc.local so that it gets run on boot, it's an unusual of starting these types of services, but it would work for both arch and artix

appeasementPolitik:
Keeping this in mind would you consider only merging the changes in `sb-volume`? It mirrors what happens when you press super+f4, but now also when you left click the module in the status bar

thetubster99:
Hey, nice feature. I found a way to work around acpid using pactl. See https://github.com/thetubster99/cookedrice/commit/23c36ba8b088b8d5549dd8f86ab7683a7b02e1d8

PR #1321, opened 2023-05-05
application/octet-stream is for arbitrary binary files

poeplva:
application/octet-stream mime-type is used for arbitrary binary files, so the best guess at opening this kind of files should be based on the extension. before this all of those were tried to be opened with zathura, now it is the last resort in case the file extension is unknown.


CheesyChocolate:
so this new config works for you? I'm getting the following error.

```bash
 --: -c: line 26: syntax error near unexpected token `)' 
```

2084x:
;; is missing after esac

poeplva:
Sorry about that, it seems like I missed the ;; while copying. I will add a new pr.

PR #1322, opened 2023-05-10
Fix File Names for Unix Environment

emrakyz:
Luke mentions about Unix naming conventions on his videos. Here is a script to increase consistency according to Unix conventions for all file names in parallel, very easily and fast in a safe way.

Luke also asks: "What do you think about naming files with underscores instead of dashes?", stating his worry about the usage of underscores seems like a "soydev" thing 😂. I give my opinion below. Actually the justification is objective compared to an opinion.

### What The Script Does

**1.** Check if the item is a directory. If so;

- **a)** Remove non-English characters.
- **b)** Replace spaces, dots, and dashes with underscores.
- **c)** Remove consecutive underscores.
- **d)** Convert the name to lowercase.
- **e)** Remove any other special characters.
- **f)** If the resulting name is empty, set it to "untitled".
- **g)** Every file or directory should start and end with an alphanumeric character.

**2.** If the item is a file, apply the same transformations as for directories, but keep the file extension intact.

**3.** Check if the original name and the new name are different. If so, and if a file or directory with the new name already exists, create a unique name.

- The script can use Dash and parallel processes, ensuring safety and performance with a subshell environment. Therefore it can even rename more than 100.000 files that have extremely weird names in 30 seconds (I have tested bash built-in functions, tr, awk and sed. None of them was faster than sed for this task, awk was very close but still slower).

**Examples of How Every File Should Look:** this_is_an_example_directory_name  **OR**  this_is_an_example_video_file.mp4

test_1

test_2

### Why "_" is Preferred Instead of a Space or a Dot or a Dash

In Unix environments, it is generally recommended to replace spaces in filenames with underscores (_), rather than dots (.) or dashes (-). This is because underscores are more commonly used and supported by Unix utilities and programming languages.

Dots (.) are typically used as a separator between a file's name and its extension, so using them to replace spaces can lead to confusion and errors. Dashes (-) are sometimes used in place of spaces, but they can be problematic because they are often used as a command-line option delimiter in Unix, which can lead to unexpected behavior.

- **Readability:** Underscores make file and directory names more readable, as they clearly separate words and components in the name, whereas spaces can be easily overlooked, and dots can be mistaken for file extensions.

- **Compatibility:** Some command line tools and scripts may not handle file names with spaces or dots properly without additional configuration or escaping. Underscores, on the other hand, do not require special handling and are generally better supported across various tools and environments.

- **URL encoding:** When sharing file paths in URLs or web applications, spaces and dots may require URL encoding (e.g., replacing spaces with "%20" and dots with "%2E"), which can make the URLs less readable and more cumbersome to work with

### The Reason Behind Using a Subshell Environment

Subshells are used in the script to isolate the execution environment of each parallel process. This isolation ensures that the processes do not interfere with each other, as they have their own separate environments, including local variables and function definitions. This separation is particularly important when running multiple processes in parallel, as it reduces the risk of race conditions and other synchronization issues.

Using subshells in the script also simplifies the process of launching parallel processes. By executing the process_item function within a subshell, the script can easily leverage the -P flag of xargs to specify the maximum number of parallel processes to run. This results in improved performance and efficiency when processing a large number of files and directories.

### The Benefit of Removing Non-English Characters

- **Compatibility:** Non-English characters can cause compatibility issues with some tools, applications, or systems that are not properly configured to handle them. By removing these characters, you reduce the risk of encountering issues related to character encoding and ensure broader compatibility across different environments.

- **Consistency:** Standardizing file and directory names by removing non-English characters can make it easier to organize, search, and manage your files. It helps maintain a consistent naming convention across your file system, which can be beneficial for both human users and automated processes.

- **Accessibility:** Using only English characters in file and directory names can improve accessibility for users who may not be familiar with non-English characters or languages. This can be particularly important in multi-user or multi-language environments where not all users might be comfortable with non-English characters.

### A Lot More Details
- **find . -depth -name '*' -print0**: This find command searches for all files and directories recursively in the current directory (.). **-depth** ensures that the directory tree is traversed depth-first, and **-name '*'** matches all items. **-print0** prints the results separated by a null character (useful for handling filenames with spaces or special characters).

- **| xargs -0 -n1 -P10 -I{} sh -c ''**: The find command output is piped (|) to xargs. The **-0** option tells xargs to expect null-terminated items. **-n1** processes one item at a time. **-P10** runs 10 parallel processes. **-I{}** sets the placeholder for input items. **sh -c '...'** runs a shell script with the given commands for each input item.

- **generate_unique_name() { }**: This is a function that generates a unique name for a file or directory. It takes three arguments: the base name, the extension (if any), and the destination path. It increments a counter and appends it to the base name until a unique name is found, then returns the unique name.

- **process_item() {  }**: This is the main function that processes a single file or directory path. It sanitizes the name and renames the item if needed.

- **[ "$item_path" = "." ] && return**: This line checks if the item path is the current directory (.). If it is, the function returns without doing anything.

- **dir_name=$(dirname "$item_path"); base_name=$(basename "$item_path")**: These commands extract the directory name and base name from the item path.

- **if [ -d "$item_path" ]; then else fi**: This conditional block checks if the item is a directory (-d) and processes it accordingly.

- **new_name=$(echo "$base_name" | sed -E "s/[^a-zA-Z0-9 _.-]+//g; s/[ .-]+/_/g; s/_+/_/g; s/^_//; s/_$//; s/(.*)/\L\1/")**: This line uses sed to sanitize the base name by removing unwanted characters, replacing spaces and periods with underscores, and converting the name to lowercase. The -E flag enables extended regular expressions. This command can't use **[:alnum:]** because it gets non-English characters that way.

- **[ -z "$new_name" ] && new_name="untitled"**: If the new name is empty, it is set to "untitled".

- **file_ext="${base_name##*.}" base_name_no_ext="${base_name%.*}"**: For files, this line extracts the file extension and the base name without the extension.

- **new_name="${new_base_name_no_ext}.${file_ext}"**: For files, this line constructs the new file name with the sanitized base name and the original file extension.

- **if [ "$base_name" != "$new_name" ]; then fi**: This conditional block checks if the original name and the new name are different.

- **[ -e "${dir_name}/${new_name}" ] && new_name=$(generate_unique_name "${new_name%.*}" "${new_name##*.}" "$dir_name")**: If the new name already exists, the generate_unique_name function is called to get a unique name.

- **mv "$item_path" "${dir_name}/${new_name}" 2>/dev/null || true**: This line moves (renames) the item to the new path with the sanitized name. If an error occurs, it is redirected to /dev/null (ignored) and the script continues executing due to the || true.

- **process_item "{}"**: This line calls the process_item function with the input item path (represented by {}) as the argument.

- **' 2>/dev/null**: This part of the script suppresses any error messages by redirecting the standard error output to /dev/null.

### DIFF
--- a/.local/bin/fixnames
+++ b/.local/bin/fixnames
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+find . -depth -name '*' -print0 | xargs -0 -n1 -P10 -I{} sh -c '
+    generate_unique_name() {
+        base_name="$1"; ext="$2"; dest_path="$3"; count=1
+        [ -z "$ext" ] && new_name="$base_name" || new_name="${base_name}.${ext}"
+        while [ -e "${dest_path}/${new_name}" ]; do
+            [ -z "$ext" ] && new_name="${base_name}_${count}" || new_name="${base_name}_${count}.${ext}"
+            count=$(( count + 1 ))
+        done
+        echo "$new_name"
+    }
+
+    process_item() {
+        item_path="$1"; [ "$item_path" = "." ] && return
+        dir_name=$(dirname "$item_path"); base_name=$(basename "$item_path")
+        if [ -d "$item_path" ]; then
+            new_name=$(echo "$base_name" | sed -E "s/[^a-zA-Z0-9 _.-]+//g; s/[ .-]+/_/g; s/_+/_/g; s/^_//; s/_$//; s/(.*)/\L\1/")
+            [ -z "$new_name" ] && new_name="untitled"
+        else
+            file_ext="${base_name##*.}"
+            base_name_no_ext="${base_name%.*}"
+            new_base_name_no_ext=$(echo "$base_name_no_ext" | sed -E "s/[^a-zA-Z0-9 _.-]+//g; s/[ .-]+/_/g; s/_+/_/g; s/^_//; s/_$//; s/(.*)/\L\1/")
+            [ -z "$new_base_name_no_ext" ] && new_base_name_no_ext="untitled"
+            new_name="${new_base_name_no_ext}.${file_ext}"
+        fi
+        if [ "$base_name" != "$new_name" ]; then
+            [ -e "${dir_name}/${new_name}" ] && new_name=$(generate_unique_name "${new_name%.*}" "${new_name##*.}" "$dir_name")
+            mv "$item_path" "${dir_name}/${new_name}" 2>/dev/null || true
+        fi
+    }
+    process_item "{}"
+' 2>/dev/null


mutageneral:
Why does it use dash explicitly? I do use dash, but you should probably just use sh so that it works on systems with other shells as /bin/sh.

Also, I'm getting this error:
```
$ ./fixnames sulinos-20201112-minimal.iso 
mv: unable to evaluate 'sulinos-20201112-minimal.iso/sulinos_20201112_minimal.iso': Not a directory
$ ./fixnames out.md
$ ls out.md sulinos-20201112-minimal.iso 
out.md  sulinos-20201112-minimal.iso
```

emrakyz:
Hi @mutageneral 

It uses **Dash** explicitly because Luke's repo right now doesn't have any **POSIX** shell and trying to run this script with /bin/sh (which is linked to **Bash** on this repo right now) can give users errors. By showing that this use Dash shows this script is POSIX, fast and simple; as well as showing the fact that you can't run this on Bash but on the other hand, you are completely right and you can change it to **/bin/sh** on your system if you use a POSIX shell as a /bin/sh replacement.

**For your problem:**
This script can't rename files alone. You need to give it a PATH that is a directory. That's why you see an error.

For example let's say I have a million files and directories under the root directory `/`

When you enter: `fixnames /` it renames ALL files and directories on your system in an extremely fast way, excluding hidden files, config files, root protected files for better safety and security. So you can't rename any files that is owned by the root (system files) or you can't rename things under .local or .config in order for things not to break.

On the other hand, sometimes you may see some unrelated harmless errors (though I tried to get rid of them as much as possible). Let's say your system is too fast and you have a threadripper CPU with 192 threads. In that case you would rename files and directories so fast that there can be some race conditions (directories renamed before files) so in this case you need to run the script for the second time for renaming the files that couldn't be renamed before.

This script is so easy to use and fast so you don't need to worry about these. Just run it and your files will be renamed. Let me give you an example:

You can give it a full PATH or a relative one:
`fixnames /home/user/pictures` or `fixnames pictures` (pictures should be a directory since we first run find command on it).

It would rename all files and directories no matter how deep the files and directories are located. Let me give you a before after practical example with a video. It will fix anything and everything instantly no matter how complex or how deep your files are. If you have millions of files as well as having a very slow machine, you may need to wait for a couple minutes though:


https://github.com/LukeSmithxyz/voidrice/assets/89175311/95a7ba97-a10d-43ab-a0d4-d6da3a0f6311

narukeh:
@emrakyz  have a look at this:

https://github.com/narukeh/dotfiles/blob/master/.local/bin/personal/ffn

maybe it will help/inspire you to make it better/more-feauters

emrakyz:
Hi! @narukeh 

Thanks for sharing your script.

Can you exclusively state which features you recommend adding, so we could discuss?

Meanwhile, let me tell you my initial aims:

I have 16TB data which I have gathered in years on an external hard drive. It had lots of nested directory structures and tons of different files. Current solutions were not performant and efficient enough so I created this tool.

At the same time, I went with experimentalism and tried to create a tool with 'autistic' level simplicity, minimalism, safety and speed. With this way, this script is more fitting to the philosophy behind this repository and it is also more explainable to the people so they easily know what they exactly do with proper explanation.

On the other hand, I tried to use universally applied rules for Unix file naming and URL conventions (alphanumeric lower-cased English and underscores). Not just Unix by the way. This convention is best suited for Windows, Linux, BSD, Mac, Solaris and the Web. No application is problematic with lower-cased alphanumeric English characters with underscores. This does not require special handling on any machine.

So, the main aim of this tool is to have as less lines as possible; to apply as less commands, pipes and checks as possible; to have no dependencies and to get rid of user interactivity completely (because there is only one universal file naming conventions) while being completely safe. This also helps increasing consistency among all files within the system. So, a user can blindly invoke `fixnames /` command without further thinking to rename all files and directories within a system completely safely, and in a very fast way. You don't need to think about config files, hidden files, root protected files, extensions, directories and all.

Let's do a direct comparison and see the different aims, advantages, disadvantages of both scripts even though they pretty much do a similar job:

**This script** is extremely efficient and performant because it only applies singular shell built-in check and single `sed -E` command without any pipes, redirection or other commands. So it only invokes `dash` and `sed`, nothing else. **Your script**, however, is aimed towards singular file handling. Even though it has a pseudo feature to rename multiple files with multiple arguments or wildcards, it can only apply within a directory. Firstly, sequential file renaming with loops is not as efficient. Secondly, it uses pipes with different commands (multiple `tr` commands and `sed`) adding extra overhead. **My script** also utilizes `xargs` to use parallelism. So it's not only efficient but also parallel. 

**This script** aims to create compactness, simplicity and minimalism. This makes it easy to read and modify. The logic of the script is extremely easy to understand. On the other hand, **your script** is harder to understand. The reader needs to delve deep and think about it properly to understand what it does.

**This script** handles both files and directories (regardless of the rate of the depth) by also handling extensions without exclusively writing different commands for them separately but only adding a parameter with a built-in shell expansion. So a single, static sed command is used for files, directories and for handling extensions. **Your script's** logic revolves around processing individual filenames. It extracts the base name, extension, and performs transformations on that filename.

**This script** handles file name conflicts in a very simple, fast way. **Your script** doesn't handle this. If a file with the intended new name already exists, your script's `mv` command might silently overwrite it, leading to data loss.

**This script** is more scalable because adding a simple sed pattern is easy. On the other hand, I don't find adding a custom pattern via flags is useful because it is prone to errors and mostly unnecessary. Therewithal, using a better tool such as `bulkrename` along with `neovim` is more appropriate for manual, custom file naming.

`This script` can easily be integrated similarly with another script. The core logic within `dash -c` could be extracted and used as a modular component within a larger shell pipeline due to its potential for stdin/stdout usage. Your script has a more specific and complicated use case.

Additionally, I don't find dry running is necessary because it's obvious. It would be lower-cased alphanumeric English separated by underscores. It's the biggest advantage of my script. No one has to care. It's completely automated, consistent and safe.

`find` command is one of the best known methods to handle file names. It's POSIX. It can find and normalize all files with all kinds of names (even with newline characters). `print0` is a very good way to use them as arguments by also separating output filenames with a null character. This is essential for handling filenames with spaces, newlines, or unusual special characters. Sorting them prevents race conditions. `-depth` also help prevent race conditions by issuing file renaming before the directory itself. By issuing find, you can get all the depth, you can exclude specific files and directories and you can redirect into a subshell to do things.

Even though I could add additional features to this, I would do so only without making it significantly bigger and more complex; without breaking its parallelism. The worst thing that can happen is that an unwanted file can be renamed in a better way to be readable by a machine. I can't think of any reason a user requires an unconventional file name. Maybe an exclusion criteria could be added or a condition to apply singular file renaming could be added above the find command but in my opinion it's completely unnecessary.

emrakyz:
@mutageneral 
@narukeh 

Hey, you can check the new version I have written in Pure C instead of using subshells, redirections, piping, external programs and all. It's much faster, more minimal, safer, and it has more features.

I have taken @narukeh's recommendations into account (better help page, dry running, individual file handling; multiple argument handling). 

[No Bullshit Instant Filename Sanitizer: Threaded & Recursive & Lightweight](https://github.com/emrakyz/sanitize_fs)

This version directly uses the newer `renameat2()` and `fstatat` functions from the Linux kernel and therefore we don't even need to use lots of different calls, checks repetitively. `renameat2()` also supports renaming things at their own place and without replacing so it's even faster, safer than `mv`.

rulatir:
Not even attempting to deaccent is no-effort, not no-bullshit.

emrakyz:
> Not even attempting to deaccent is no-effort, not no-bullshit.

@rulatir I would definitely accept pull requests if you can come up with a robust, universally acceptable solution with your valuable effort.

RetroViking:
@emrakyz , if using `sort` and `xargs` is not strictly necessary for you, a version like this could be faster (not as fast as your C version though):

```
#!/bin/dash

find "${1}" -depth \( -path '*/.*' -o -path '*/.*/*' -o -path '.' \) -prune -o \( -type f -o -type d \) -exec sh -c '
for file do
	base="${file##*/}"
	path="${file%/*}"

	pattern="s/[^a-zA-Z0-9 ._-]//g; s/[ .-]/_/g; s/_+/_/g; s/^_+//; s/_+$//; s/[A-Z]/\L&/g"

	[ -f "${file}" ] && pattern="${pattern}; s/_([^_]+)$/.\\1/"

	clean_name="$(printf "%s\n" "${base}" | sed -E "${pattern}")"
	new_name="${clean_name}"
	counter=1

	while [ "${base}" != "${new_name}" ] && [ -e "${path}/${new_name}" ]; do
		new_name="${clean_name}_${counter}"
		counter=$((counter + 1))
	done

	[ "${base}" != "${new_name}" ] && mv -- "${file}" "${path}/${new_name}"
done
' sh {} +
```

RetroViking:
@emrakyz 

On my system this is faster, by using `&` instead of `xargs`. However, it does not do sorting.

```
#!/bin/sh

find "${1}" -depth \( -path '*/.*' -o -path '*/.*/*' -o -path '.' \) -prune -o \( -type f -o -type d \) -exec sh -c '
for file do
    {
        base="${file##*/}"
        path="${file%/*}"

        pattern="s/[^a-zA-Z0-9 ._-]//g; s/[ .-]/_/g; s/_+/_/g; s/^_+//; s/_+$//; s/[A-Z]/\L&/g"

        [ -f "${file}" ] && pattern="${pattern}; s/_([^_]+)$/.\\1/"

        clean_name="$(printf "%s\n" "${base}" | sed -E "${pattern}")"
        new_name="${clean_name}"
        counter=1

        while [ "${base}" != "${new_name}" ] && [ -e "${path}/${new_name}" ]; do
            new_name="${counter}_${clean_name}"
            counter=$((counter + 1))
        done

        [ "${base}" != "${new_name}" ] && echo mv -- "${file}" "${path}/${new_name}"
    } &
done
wait
' sh {} +
```

emrakyz:
@RetroViking I haven't been using it for a while :)

C version is much more minimal, safer and especially faster.

Though I made it even safer/faster by porting it to Rust. Maybe I'll also find a convenient way to de-accent. It's simply instant for even a million of nested directories.

PR #1327, opened 2023-05-13
weath: Add option to get forecast from a different location

appeasementPolitik:
You can now add the forecast location as a parameter to weath, very useful if you want to quickly know what the weather is like somewhere else. Had to use a function because aliases don't support command line parameters.

`weath`
`weath 'Rome'`

Also added the 'cp' option so you can copy the forecast as plaintext to your clipboard, so that it can be shared. I've found myself using screenshots for it before, but that's impractical.

`weath cp`
`weath cp 'Rome'`

### DIFF
--- a/.config/shell/aliasrc
+++ b/.config/shell/aliasrc
@@ -20,7 +20,7 @@ se() {
 
 weath() {
 	[ -z "$1" ] && less -S ${XDG_CACHE_HOME:-$HOME/.cache}/weatherreport ||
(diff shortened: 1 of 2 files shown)


TheYellowArchitect:
Those changes are great, but I think it should call a `.local/bin` bash script (containing your command/function), because aliasrc isnt meant to hold logic or big commands

appeasementPolitik:
Thanks. You're right, I've now made it a separate script.

PR #1330, opened 2023-05-24
bulkrename using built-ins instead of using a completely separate program

mhdna:


emrakyz:
I find your approach really good and brilliant. I have made some improvements, but I won't open a new PR for it since the idea is yours. Please review it, maybe we could improve it more.

- My version is much simpler and minimal.
- Your script uses a variable $fs to capture the output of ls command. However, it is unclear why this is necessary because $fs is only used once and there is no apparent need for it.
- Your script exits without removing temporary files.
- My script uses the read command with a tab as the field separator, which is a simpler and more efficient approach. Your script uses printf and cut to separate the old and new names, which is less straightforward and requires more processing.
- My script processes the renaming of files sequentially. This might be easier to understand for someone reading the code.

```
cmd bulkrename ${{
tmpfile_old=$(mktemp)
tmpfile_new=$(mktemp)
ls > "$tmpfile_old"
cp "$tmpfile_old" "$tmpfile_new"
nvim "$tmpfile_new"
[ $(ls | wc -l) -eq $(wc -l < "$tmpfile_new") ] || { rm -f "$tmpfile_old" "$tmpfile_new"; exit 1; }
paste "$tmpfile_old" "$tmpfile_new" | while IFS="$(printf '\t')" read -r oldname newname
do
  [ "$oldname" = "$newname" ] || { [ -e "$newname" ] || mv -- "$oldname" "$newname"; }
done
rm -- "$tmpfile_old" "$tmpfile_new"
}}
```

mhdna:
@emrakyz 
For the sake of clarity, I'd like you to note that this snippet of code was taken from the LF docs  with a little modification for my personal use, without me thinking much about enhancing it.

Thanks for pointing out, your code does in fact look more efficient and straightforward.

A couple of things I'd like to be noticed, however
- `$fs` isn't put for no reason. It's a global variable that LF initializes by default for currently selected files. So, we only assign it locally to `$(ls)` if there were no files selected (i.e. If `$fs` is empty).
- I noticed that unlike `$(ls)`, `$fs` returns the full path for selected files, so I'll put back the `$(basename -a $fs)` statement.
- We could assign IFS to `$'\t'` instead of calling `$(printf '\t')`.
- `rm --` isn't needed for removing the tmp files, since by default, files created using `$(mktemp)` always start with tmp.XXX.
- Since echo usually runs faster as a built-in command of shell, I replaced printf with it.

Combining the two snippets, this might be the end result:

```
cmd bulkrename ${{
    tmpfile_old="$(mktemp)"
    tmpfile_new="$(mktemp)"
    if [ -n "$fs" ]; then
        fs="$(basename -a $fs)"
    else
        fs="$(ls)"
    fi
    echo "$fs" > "$tmpfile_old"
    echo "$fs" > "$tmpfile_new"
    $EDITOR "$tmpfile_new"
    [ "$(wc -l < "$tmpfile_old")" -eq "$(wc -l < "$tmpfile_new")" ] || { rm -f "$tmpfile_old" "$tmpfile_new"; exit 1; }
    paste "$tmpfile_old" "$tmpfile_new" | while IFS=$'\t' read -r src dst
    do
        [ "$src" = "$dst" ] || [ -e "$dst" ] || mv -- "$src" "$dst"
    done
    rm -f "$tmpfile_old" "$tmpfile_new"
    lf -remote "send $id unselect"
}}
```

Any other suggestions?

emrakyz:
@mhdna 

Thanks for the response. I have made more improvements and clarify one of your supposed changes:

- **IFS=$'\t' does not work for all shells**

The $'\t' syntax is an example of ANSI-C quoting, supported in some shells (like bash and zsh) but not all.

The dash shell (very popular, faster and minimal one) for example does not support ANSI-C quoting. That's why **IFS=$'\t'** doesn't work in your script.

**IFS="$(printf '\t')"** works because the printf command is a standalone program that interprets \t as a tab, and its output (a tab character) is assigned to IFS. This is a more portable way to set IFS to a tab character, as it does not rely on features specific to certain shells.

- We can remove if statements and use logical operators (Also Luke's preferred way).

- Using echo is generally ok. It can stay the same. But there can be problems for edge cases with "ls" and "echo" usage. Some file names would be problematic with them. So using a shell glob (*) in a loop with also using printf instead of echo would be a more robust solution to cover the edge cases but it's not completely necessary for the 99% of people's use cases and it could make the script more complex than it already is. 

Here is the latest script:
```
cmd bulkrename ${{
tmpfile_old="$(mktemp)"
tmpfile_new="$(mktemp)"

[ -n "$fs" ] && fs=$(basename -a $fs) || fs=$(ls)

echo "$fs" > "$tmpfile_old"
echo "$fs" > "$tmpfile_new"
$EDITOR "$tmpfile_new"

[ "$(wc -l < "$tmpfile_old")" -eq "$(wc -l < "$tmpfile_new")" ] || { rm -f "$tmpfile_old" "$tmpfile_new"; exit 1; }

paste "$tmpfile_old" "$tmpfile_new" | while IFS="$(printf '\t')" read -r src dst
do
    [ "$src" = "$dst" ] || [ -e "$dst" ] || mv -- "$src" "$dst"
done

rm -f "$tmpfile_old" "$tmpfile_new"
lf -remote "send $id unselect"
}}
```

mhdna:
@emrakyz
Alright. I've committed these changes.
Thanks!

PR #1333, opened 2023-05-25
Re-add whitespace to some buffer types as needed.

equwal:
Very arcane.

C ANSI standard requires a newline at the end. GCC will complain.[1]

Email signatures should have a space after the -- according to the standard. Neomutt does this automatically for you if you have a signature.[2]

[1]: It's in the ANSI standard pdf somewhere. Can't be bothered to find it right now. Think it is in c99 too.
[2]: https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter


equwal:
Big thanks to who pointed this out to me in a personal email.

PR #1337, opened 2023-05-27
Corrects "application/octet-stream is for arbitrary binary files"

poeplva:
#1321 had a syntax error, this commit corrects it.


2084x:
might be worth adding a line for transport streams as well:

`ts) setsid -f mpv $f -quiet >/dev/null 2>&1 ;;`

for some reason they are recognised as application/octet-stream instead of video/mp2t

PR #1340, opened 2023-06-21
Launch dwm in a dbus session

LalleSX:
This makes it so programs can "talk" to each-other for example a problem that i ran into was with gnome-keyring not being able to be used in programs like in vscodium when getting my ssh keys

While the init system runs dbus, it is  only the system wide bus, not a session bus

### DIFF
--- a/.config/x11/xinitrc
+++ b/.config/x11/xinitrc
@@ -14,4 +14,4 @@ else
 	. "$HOME/.xprofile"
 fi
 
-ssh-agent dwm
+exec dbus-launch --exit-with-session dwm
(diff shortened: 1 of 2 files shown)


hooregi:
I haven't had a chance to properly debug this on my end, as it might just be an isolated issue. But I tried this configuration and the session ends up freezing at -apparent- random points in time. Only rebooting will take me out of it. The thing is, the session is still active, since if a video is playing then it'll keep playing. But keymaps and other peripherals won't get registered.

LalleSX:
@hooregi I have updated it to what is currently my xinitrc, keymaps & stability work fine on my end
I tested it out not that long ago on a fairly new system where it fixed the bug with VSCode not being able to "talk" with gnome keyring

hooregi:
Yeah, after a few boots and about 11 hours of uptime there seem to be no issues. I'll obviously keep testing it; wonder why it failed with the previous config, my guess is that it was missing the `dbus-update-activation-environment --all` line since it's needed for gnome-keyring outside of usual desktop environments as per the [arch wiki](https://wiki.archlinux.org/title/GNOME/Keyring).

PR #1344, opened 2023-07-08
Fixed formatting bug with notify-send

ghalv:


mhdna:
Make sure to add a `\<` and a `\>` before and after `$(date +'%e')`, otherwise you'd get something like this:

ghalv:
Thank you, implemented your suggestion!

PR #1355, opened 2023-08-28
timeout forecase and moonphase at one second

josephkerkhof:
If there is a network disruption or downtime for the `wttr.in` service, the `curl` command should timeout. If the `curl` command does not have a timeout, it would prevent other dwmblocks from being updated while the `sb-forecast` or `sb-moonphase` hangs.

I chose the timeout to be 1 second to be consistent with other status bar scripts like `sb-price` and `sb-iplocate`. One second also makes sense because some dwmblock scripts update every second.


appeasementPolitik:
Also look at this PR, I've solved it by doing the curl downloading in the background so the rest of the statusbar can run instantly:

https://github.com/LukeSmithxyz/voidrice/pull/1296

PR #1360, opened 2023-09-08
lf-scope: adds SVG preview creating cache.

lpanebr:
This uses inkscape (v1.3) to convert. It's quick and does not open the GUI.


matthew55:
Is there any reason to use Inkscape to accomplish this? I am able to get seemingly the same results just using convert.

```bash
image/svg+xml)
	CACHE="${XDG_CACHE_HOME:-$HOME/.cache}/lf/thumb.$(stat --printf '%n\0%i\0%F\0%s\0%W\0%Y' -- "$(readlink -f "$1")" | sha256sum | cut -d' ' -f1)"
	[ ! -f "$CACHE" ] && convert "$1" "$CACHE.jpg"
	image "$CACHE.jpg" "$2" "$3" "$4" "$5" "$1" ;;
```

lpanebr:
`convert` will export using the document's page and may result in cropping:
image

**inkscape** exports the objects:
image

Of course, depending on the art exporting objects could algo give undesired results...

Since I have both and svg files download from the web usually won't have other objects laying around I chose to use inkscape.

matthew55:
Thanks for the response, I didn't realize convert could cause cropping issues.

PR #1365, opened 2023-10-17
fixed selecting Romanian language

tragdate:
The script did not actually select Romanian language when picked, it was basically US layout


xslendix:
no.

PR #1368, opened 2023-11-03
POSIX compliant shell profile

thirtysixpw:
Shell profile is placed in `~/.config/shell/profile` along side shell-agnostic files such as `aliasrc`. Profile file in its turn has zsh shebang and zshisms. This PR replaces them with basic coreutils.

Also pleased shellcheck and removed variable fallbacks, that are unnecessary, because variables are defined in the same file; removed outdated docs.

### DIFF
--- a/.config/shell/profile
+++ b/.config/shell/profile
@@ -1,14 +1,12 @@
-#!/bin/zsh
+#!/bin/sh
+# shellcheck disable=SC2155
 
-# profile file. Runs on login. Environmental variables are set here.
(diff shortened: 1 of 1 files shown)


thirtysixpw:
I know what you [don't like](https://github.com/LukeSmithxyz/voidrice/pull/313#issuecomment-497974109) `find` and have have rejected pull requests similar to this (#239, #313) in past. But since then you have add `find` in `se` shell function and other scripts.

PR #1369, opened 2023-11-03
Fix pdf and plain text previws in lf

thirtysixpw:
Pdf previws don't appear because are generated with wrong filename.

Plain text previws with `bat` don't respect terminal theme and have annoying border.

### DIFF
--- a/.config/lf/scope
+++ b/.config/lf/scope
@@ -13,10 +13,6 @@ image() {
 	fi
 }
 
-ifub() {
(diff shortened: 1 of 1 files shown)


thirtysixpw:
Yes, `pdftoppm` worked fine, I probably messed something up in another place. Mine is on 23.10.0 version too.

thirtysixpw:
Reverted pdf filename, now it works fine.

This pull request also fixes plain text previews, done with `bat`, forcing them to use terminal theme.

PR #1374, opened 2023-11-18
lfrc - Added hotkey which converts selected videos (webm,mkv,mp4) to mp3 via ffmpeg

TheYellowArchitect:
It ignores non-videos and videos which have no audio layer ([ffprobe](https://ffmpeg.org/ffprobe.html#Description) ftw). Moves them to the music folder, and notifies the user.

Change whatever you wish, as long the functionality remains the same (converting a song which is .webm ftw)

### DIFF
--- a/.config/lf/lfrc
+++ b/.config/lf/lfrc
@@ -130,12 +130,11 @@ cmd bulkrename ${{
     lf -remote "send $id unselect"
 }}
 
-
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
This seems like the kind of thing that should be a separate shell script, *perhaps* runable in lf with a shortcut, same with #1375.

PR #1378, opened 2023-11-23
Simplify lfcd

Nawar9:
lf recently added the `-print-last-dir` flag, eliminating the need for a temporary file in the `lfcd` function


2084x:
ueberzug previews stop working if you prefix `lf` with `command` btw.

try opening st, hitting ctrl+o, then navigating to an image.

PR #1382, opened 2023-12-18
Update emoji set

catfromplan9:


narukeh:
bloat included. i say to deny this rn, so that we dont even have to look at it.

catfromplan9:
> bloat included. i say to deny this rn, so that we dont even have to look at it.

Yeah, its every emoji

I need the transgender flag emoji, it is very important to me to have it, and rn as is it is not available

PR #1384, opened 2023-12-22
dmenuunicode script simplify

sayanta01:
getting emoji using curl.


TheYellowArchitect:
Please do not be insulted by what I am about to write: I suggest this to be rejected because it depends on the internet (github specifically) and so it doesn't work all the time (what if I use my laptop in my log cabin?), and also because it removes comments which help script newbies like me to understand the basics in order to delve later in more advanced syntax

PR #1391, opened 2024-01-14
Update sb-internet

btkoch:
Fix ethernet icon for systems with multiple NICs wherein `cat /sys/class/net/e*/operstate ` outputs multiple lines. Icon now shows connected when one or more NICs are connected.

### DIFF
--- a/.local/bin/statusbar/sb-internet
+++ b/.local/bin/statusbar/sb-internet
@@ -25,7 +25,7 @@ elif [ "$(cat /sys/class/net/w*/operstate 2>/dev/null)" = 'down' ] ; then
 fi
 
 # Ethernet
-[ "$(cat /sys/class/net/e*/operstate 2>/dev/null)" = 'up' ] && ethericon="🌐" || ethericon="❎"
+[ "$(cat /sys/class/net/e*/operstate 2>/dev/null | grep "up")" = 'up' ] && ethericon="🌐" || ethericon="❎"
 
 # TUN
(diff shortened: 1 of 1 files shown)


mutageneral:
This does not work if 2 NICs are "up" and grep return both, and then both "up" then gets compared to a single "up".
You can try the following for proof:
```bash
$ [ "$(echo -e "up\nup" | grep "up")" = 'up' ] && echo "🌐" || echo "❎"
❎
```

This would be better instead:
```bash
cat /sys/class/net/e*/operstate 2>/dev/null | grep -q "up" && ethericon="🌐" || ethericon="❎"
```
If grep finds at least one instance of "up" it will return a "0" (no errors) return code and "🌐" will be put in the variable.
`-q` tells grep to be quiet, just like piping to /dev/null but better.
If grep does not find at least one instance of "up" it will return "1" (errors) return code and "❎" will be put in the variable.

You can try the following for proof:
```bash
$ echo -e "up\nup" | grep -q "up" && echo "🌐" || echo "❎"
🌐
```

btkoch:
Thanks for pointing that out. This is why I rarely open PRs.

thetubster99:
Small suggestion, you can change

```
cat /sys/class/net/e*/operstate 2>/dev/null | grep -q "up" /sys/class/net/e*/operstate && ethericon="🌐" || ethericon="❎"
```
to:
```
grep -q -m1 'up' /sys/class/net/e*/operstate && ethericon="🌐" || ethericon="❎"
```
It's slightly faster because cat is not needed anymore, and the -m1 option makes grep instantly stop after it has found its first match.

mutageneral:
holy shit i was catting into grep xd

PR #1396, opened 2024-02-04
shortcuts: zathura full screen mode shortcuts should work as before

visrut7:
- Issue
  - After making Zathura full screen shortcuts were not working as before

- Platform
  - Debian based


2084x:
[wal template](https://github.com/LukeSmithxyz/voidrice/blob/master/.config/wal/templates/zathurarc) should be updated as well

PR #1399, opened 2024-03-04
Removing the need of ~/.zprofile

narukeh:
When we have this, and `/etc/zsh/zshenv` containing: ZDOTDIR="${XDG_CONFIG_HOME:-$HOME/.config}/zsh"
then we will not need ~/.zprofile

### DIFF
--- a/.config/zsh/.zprofile
+++ b/.config/zsh/.zprofile
@@ -1,0 +1,1 @@
+../shell/profile


.config/zsh/.zprofile
```diff
@@ -0,0 +1 @@
+../shell/profile
```
GhoulBoii, line 1:
```suggestion
source "${XDG_CONFIG_HOME:-$HOME/.config}/shell/profile"
```
I get a permission denied error and to solve this I source it using its path.

PR #1402, opened 2024-03-26
Update xinitrc

ewof:
https://codeberg.org/librewolf/issues/issues/1745

fixes issue with librewolf (and probably other dbus applications)

### DIFF
--- a/.config/x11/xinitrc
+++ b/.config/x11/xinitrc
@@ -15,4 +15,4 @@ else
 fi
 # Activate dbus variables
 dbus-update-activation-environment --all
-ssh-agent dwm
+dbus-launch ssh-agent dwm


BEST8OY:
@LukeSmithxyz 
This causes issue with some other stuff
https://bbs.archlinux.org/viewtopic.php?pid=2165005

v3natio:
@BEST8OY @ewof

The following can be used for machines that use `systemd` and require the user environment variables for the `dbus` session.

```sh
eval $(dbus-launch --sh-syntax) # starts dbus daemon for the session
systemctl --user import-environment DBUS_SESSION_BUS_ADDRESS DISPLAY XAUTHORITY # updates systemd user env
eval $(ssh-agent) # starts ssh-agent
systemctl --user import-environment SSH_AUTH_SOCK SSH_AGENT_PID # updates systemd user env
exec dwm # starts dwm
```

PR #1406, opened 2024-05-03
Create sb-bghitness, show brightness in the bar.

ArgusGuardian:
script to add brightness to the dwmblocks.


appeasementPolitik:
Would something like `cat /sys/class/backlight/*/brightness` not be better than `brightnessctl`? It doesn't require extra programs

ArgusGuardian:
> Would something like `cat /sys/class/backlight/*/brightness` not be better than `brightnessctl`? It doesn't require extra programs

that would be the more minimalist way of doing it, thanks for pointing that out.

PR #1415, opened 2024-05-23
Add OCR to maimpick

ZmoleCristian:
Sometimes you just need to grab some text from a video


ZmoleCristian:
Welcome back @LukeSmithxyz
Thanks for fixing LARBS, makes my job easier to allow my followers to move to a based workflow!

PR #1416, opened 2024-05-23
add select option for OCR

ZmoleCristian:
oopsie!


2084x:
should also change the dmenu  command to `-l 7`

PR #1418, opened 2024-06-01
Enable wrap-around scrolling

public-void-ryan:
It was getting annoying having to scroll up and down a big list of files or directories. LF has a built-in wrap around function


LukeSmithxyz:
I think `g` and `G` suffice.

PR #1429, opened 2024-08-03
Make maimpick OCR to work with languages other than English.

forgoty:
If you work with human languages other than English this might be useful. List of available languages for OCR is selected from local tesseract configuration.


emrakyz:
Your command is a little bit bloated. You don't need to use several processes. Especially try to never use `awk`, which is an extremely big and complex software for a simple task. `awk` is only necessary for extremely complex tasks (sometimes it makes things easier though).

You can do the same thing with a single sed command.

```
tesseract --list-langs 2>/dev/null | sed '1d;/^List/d;/osd/d'
```

On the other hand, `tesseract` can work with stdin and stdout. You don't need to create a temp image. You can remove `ocr_cmd` function completely.

```
maim -u -s | tesseract stdin stdout -l eng | xclip -sel clip
```

By the way your case statements lack readability. You should use a printf abstraction.

```
opts="a selected area
current window
full screen
a selected area (copy)
current window (copy)
full screen (copy)
copy selected image to text
copy selected image to text (choose language)"

opt="$(printf "%s\n" "${opts}" | dmenu -l 7 -i -p "Screenshot which area?")"

case "${opt}" in
        "a selected area") command ;;
        ...) command ;;
esac
```

forgoty:
@emrakyz thanks for useful comment. I've adjusted the script. Please take a look.

emrakyz:
@forgoty 

Looks good.

Keep that in mind, if you ever want to use the same option list in an indented way (inside functions), you can use `cat <<- EOF`. But this only works with tabs, not spaces.

Such as:

```
main() {
        C="$(
                cat <<- EOF | m "Torrents"
	List
	Add
	Remove
	Torrent Prio
	File Prio
	Start/Stop
	Disable All Files
	Search
	Killall
	EOF
        )"

        case "${C}" in
                "Add") add ;;
                "Remove") remove ;;
                "List") list ;;
                "Torrent Prio") tor_prio ;;
                "File Prio") file_prio ;;
                "Start/Stop") start_stop ;;
                "Disable All Files") no_download ;;
                "Search") search ;;
                "Killall") killsv ;;
                *) exit ;;
        esac
}
```

PR #1431, opened 2024-08-10
Change nvim background color to dark mode.

triaseptmbr:
Maybe it's just personal preference, I think it looks better when set to dark mode. It also matches to default theme.

Dark Mode
pic-window-240810-1459-33

Light Mode
pic-window-240810-1500-03


lukerb52:
This was one of the changes I made in my PR #1465 , which overhauls the neovim configuration by using the new init.lua

PR #1433, opened 2024-09-02
aliasrc  |  improve se()

emrakyz:
1. Remove external commands like `find`.
2. Remove extensions (possibly `.sh`) and path (`/home/username/.local/bin/`) from the names in `fzf`.
3. Only open the editor if there is a selection.

Do all of these without using `find`, `sed`, `grep`.

- First line creates an array with the files in the scripts directory.

- Second line removes path `:t` and the extensions `:r` from the scripts.

- `[[ "${c}" ]]` checks if this variable is non-empty.

- `${${(M)s:#*/${c}*}[1]} ` This works like `grep`

`(M)` enables "match" mode.
`:#` anchors the pattern to the start of each array element.
`*/${c}*` matches any path containing the selected basename.
`[1]` selects the first matching item.

### DIFF
--- a/.config/shell/aliasrc
+++ b/.config/shell/aliasrc
@@ -14,7 +14,7 @@ for command in mount umount sv pacman updatedb su shutdown poweroff reboot ; do
 done; unset command
 
 se() {
-	s=("${HOME}/.local/bin/"*)
+	s=("${HOME}/.local/bin/"**/*(.))
 	c="$(print -lnr ${s:t:r} | fzf)"
 	[[ "${c}" ]] && "${EDITOR}" ${${(M)s:#*/${c}*}[1]}
 }
--- a/.config/shell/aliasrc
+++ b/.config/shell/aliasrc
@@ -14,9 +14,10 @@ for command in mount umount sv pacman updatedb su shutdown poweroff reboot ; do
 done; unset command
 
 se() {
-	choice="$(find ~/.local/bin -mindepth 1 -printf '%P\n' | fzf)"
-	[ -f "$HOME/.local/bin/$choice" ] && $EDITOR "$HOME/.local/bin/$choice"
-	}
+	s=("${HOME}/.local/bin/"*)
+	c="$(print -lnr ${s:t:r} | fzf)"
+	[[ "${c}" ]] && "${EDITOR}" ${${(M)s:#*/${c}*}[1]}
+}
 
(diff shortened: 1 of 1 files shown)


2084x:
doesn't pick up scripts in subdirectories eg. statusbar/

emrakyz:
@2084x 

We can use a more capturing glob pattern then that is supported by Zsh.

`s=("${HOME}/.local/bin/"**/*(.))`

This would capture sub-directories' contents too.

.config/shell/aliasrc
```diff
@@ -14,9 +14,10 @@ for command in mount umount sv pacman updatedb su shutdown poweroff reboot ; do
 done; unset command
 
 se() {
-	choice="$(find ~/.local/bin -mindepth 1 -printf '%P\n' | fzf)"
-	[ -f "$HOME/.local/bin/$choice" ] && $EDITOR "$HOME/.local/bin/$choice"
-	}
+	s=("${HOME}/.local/bin/"**/*(.))
+	c="$(print -lnr ${s:t:r} | fzf)"
+	[[ "${c}" ]] && "${EDITOR}" ${${(M)s:#*/${c}*}[1]}
```
RetroViking, line 19:
```suggestion
	[[ -z "${c}" ]] && return 1
	[[ "${c}" ]] && "${EDITOR}" ${${(M)s:#*/${c}*}[1]}
```
For purely aesthetic purposes. Presents that error which appears when you press `esc` instead of selecting something.

Very cool. The abbreviated variable names (`s`, `c`) might be confusing for a noob, especially since there's no comment on what the function does. On the other hand, the naming scheme is consistent with a minimalist aesthetic which I dig.

emrakyz: Writing the variable directly in brackets have the same meaning on all shells.

`[[ -n "${VARIABLE}" ]]`
`[[ "${VARIABLE}" ]]`

are the same things.

Two brackets are substantially faster but won't work in Dash/sh. Though we already have zsh here.

RetroViking: When I first tried this it would send an empty string if I pressed `esc` instead of making a selection. Adding either the `[[ -n ${var} ]]` test or `[[ -z ${var} ]] && return 1` would fix it. After reading your reply I tried the original version again and it seems to work now. This could've been related to some configuration I had back then, although I can't identify which.

PR #1437, opened 2024-10-22
Fix dangerous copy/move script in lf

futureisfoss:
Context: I accidentally pressed the C key on my .config directory and was presented with a list of directories to copy it to, then I pressed escape to quit the fzf menu without choosing anything - instead of doing nothing the script copied all of the contents inside my .config directory into my home directory. After dealing with that mess I decided to make this PR which does the following:

- Allow users to escape out of the fzf menu without unexpected copies
- Asks the user for confirmation before copying/moving files
- Some improvements in the UI

### DIFF
--- a/.config/lf/lfrc
+++ b/.config/lf/lfrc
@@ -92,25 +92,37 @@ cmd delete ${{
 }}
 
 cmd moveto ${{
-	clear; tput cup $(($(tput lines)/3)); tput bold
 	set -f
-	clear; echo "Move to where?"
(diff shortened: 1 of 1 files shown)


RetroViking:
Thank you for the pull request. I'd suggest adding a `clear` command somewhere at the beginning of the `moveto` and `copyto` functions in order to avoid leftover messages like these:

pic-window-241022-2137-16

futureisfoss:
@RetroViking Thanks for reporting the bug, I have made a new PR for a quick fix.

On larbs there is this line in the `.config/shell/profile` file:
```
export FZF_DEFAULT_OPTS="--layout=reverse --height 40%"
```
I had that above line commented out on my system, which is why this bug went unnoticed while testing.

PR #1442, opened 2024-11-09
Removes the arkenfox-auto-update script

### LINKED ISSUE #1439: Delete Arkenfox script hook
Hello
There is a hook that runs after updating firefox, librewolf.
Based on what the maintainer of the package on AUR says it's pointless to run arkenfox updater after each update!
image

aartoni:
According to the findings in #1439, the arkenfox-auto-update script is neither working nor needed. Closes #1439.


aartoni:
There's a [companion PR](https://github.com/LukeSmithxyz/voidrice/pull/1442) that should be merged first to avoid broken LARBS deployments.

BEST8OY:
@aartoni 
What do you think on triggering the script on '''arkenfox-user.js-git''' and '''arkenfox-user.js'''

aartoni:
I'd prefer the hook to ship with the packages you've mentioned. I'm trying to get in touch with the packager for that.

I believe that this makes sense because if I'm having pacman handling the `user.js` for me, then I also want it to keep it up to date. @BEST8OY What do you think?

BEST8OY:
> What do you think?

That would be the ideal way.

PR #1443, opened 2024-11-09
Latexmk

aartoni:
Removes `texclear` and simplifies `compiler` by using `latexmk`. Compared to the previous version, this `latexmk` configuration:

- sets the default format to XeLaTeX (`$pdf_mode = 5`);
- compiles in silent mode; errors can still be found in the log file;
- autodetects Biber/BibTeX, using Biber for BibLaTeX;
- calls Biber/BibTeX only if the bib files exist, reducing compiler calls by one;
- outputs all compiled artifacts to `out`, simply add that to your `.gitignore` instead of listing numerous files;
- simplifies configuration overwriting by honoring the `.latexmkrc` file from the root of any LaTeX project; `VimLeave` and `compile` will read those.
- Automatically detects compiled artifacts to delete during cleanup, thanks to `$cleanup_includes_cusdep_generated` and `$cleanup_includes_generated` (goodbye `texclear`).

### DIFF
--- a/.config/latexmk/latexmkrc
+++ b/.config/latexmk/latexmkrc
@@ -1,0 +1,10 @@
+$bibtex_use = 1.5;
+$cleanup_includes_cusdep_generated = 1;
+$cleanup_includes_generated = 1;
+$out_dir = "out";
+$pdf_mode = 5;
+$silent = 1;
+
+# SyncTeX
+push(@generated_exts, ("synctex.*"));
+push(@extra_xelatex_options, '-synctex=1') ;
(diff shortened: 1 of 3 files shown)


aartoni:
This PR had already been pushed before (#1428), however I've mistakenly closed it by deleting the branch on my fork of LARBS. Anyway, this one has an improved description.

v3natio:
@aartoni correct me if I'm wrong, but I believe `$out_dir = "out";` should be replaced with `$aux_dir = "out";` so that the output file lives in the same directory of the project file instead of inside `\out`.

aartoni:
@v3natio I usually keep all the compilation artifacts in `out`, including `.pdf`s. Is there any feature in particular you're missing after this change?

v3natio:
> @v3natio I usually keep all the compilation artifacts in `out`, including `.pdf`s. Is there any feature in particular you're missing after this change?

Well I'm not a user of Luke's config, although I do take inspiration from the dotfiles. But I just tested it and I think that `opout` won't work since it expects the output file to live in the root of the project.

PR #1444, opened 2024-11-09
Group multi-process program usage

xsmh:
Programs like Chrome run multiple processes that take up the entire notification window. This commit solves this issue by grouping and aggregating CPU and memory usage for multi-process programs into single entries.


LukeSmithxyz:
Very nice!

PR #1445, opened 2024-11-13
Add low battery notification

xsmh:
- Make a flag file to track if a low battery notification has been sent.
- Send a critical notification if the battery is discharging and has less than 25% capacity.
- Remove the flag file when the battery status changes from discharging to another state.

### DIFF
--- a/.local/bin/statusbar/sb-battery
+++ b/.local/bin/statusbar/sb-battery
@@ -16,6 +16,9 @@ case $BLOCK_BUTTON in
 	6) setsid -f "$TERMINAL" -e "$EDITOR" "$0" ;;
 esac
 
+# Flag file to track if low battery notification has been sent
+NOTIFY_FLAG="/tmp/low_battery_notified"
+
 # Loop through all attached batteries and format the info
 for battery in /sys/class/power_supply/BAT?*; do
 	# If non-first battery, print a space separator.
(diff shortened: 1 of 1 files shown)


sudo-Tiz:
why does your script need a file in /tmp ?

Here is what i'll do instead:

```bash
[ "$status" = "🔋" ] && [ "$capacity" -le 10 ] && (notify-send "🔋❗ Battery low 🔋❗" "Please charge the computer ASAP"; ring /home/$USER/.local/share/larbs/ringtone/low-battery.mp3 >/dev/null 2>&1)

[ "$status" = "🔋" ] && [ "$capacity" -le 5 ] && (ring /home/$USER/.local/share/larbs/ringtone/wasted.mp3 >/dev/null 2>&1 ; sudo systemctl hibernate)
```

ShahramMohammed:
@sudo-Tiz because it is inside a loop and it will keep sending the notification over and over without that file

aartoni:
@ShahramMohammed this used to be a feature, I remember having it a few years ago. You might be able to find the original script by looking at the commit history. I was wondering whether [this was the script](https://github.com/LukeSmithxyz/voidrice/blob/756ac03961513d1baa6d2673fa26ecf2c87f9c6b/.local/bin/cron/cronbat), however I don't remember having to enable it manually.

ShahramMohammed:
@aartoni why was it removed?

aartoni:
I've asked @LukeSmithxyz myself in the past but [haven't received any answer](https://github.com/LukeSmithxyz/voidrice/commit/f47e40243f2774f68bfa02ebee9289b0ce26564b#commitcomment-100458355).

LukeSmithxyz:
Yeah, we used to have this, but I removed it as unnecessary. You can add it to your own system if you want, but I find it superfluous as default especially considering the presence of the battery icon.

PR #1449, opened 2025-01-25
Fix ImageMagick-related actions in nsxiv

aartoni:


RetroViking:
Thanks for the PR. The only downside is that the thumbnails in `lf` don't update to match the modifications performed with `magick`.

By the way, the file operations (copy and move) are totally broken when the selected destination directory (bookmark) is a parameter substitution like `${XDG_DOWNLOAD_DIR:-$HOME/Downloads}` (iirc even `$HOME` by itself has this issue). This is because `"$destdir` needs to be expanded before the `[ ! -d "$destdir" ]` test is performed, like so:
```
[ -n "$destdir" ] && destdir=$(eval printf %s\\n "$destdir")
```
Note also that `sed "s|~|$HOME|g"` (used in setting `$destdir`) does not actually do the `~` -> `$HOME` substitution, and is redundant anyway since `$destdir` needs to be expanded afterwards.

Also, a bit of a nitpick:
```
printf "%s" "$file" | tr -d '\n' | xclip -selection clipboard &&
```
The `printf "%s" "$file"` part already outputs the content without a trailing newline, so `tr -d '\n'` does nothing.

aartoni:
Thanks for pointing that out! That script can definitely benefit from the improvements you mentioned, I'd say it's better to open an issue to choose how to move forward. To be honest I'm not even sure that the average use case involves copying/moving pictures outside of `~pp` but I'd prefer to discuss this on the new issue.

@RetroViking if you can please open the issue also remember to mention this PR to keep track of our conversation and notify any account following this thread.

PR #1450, opened 2025-01-27
Fix the J keybinding in lf

futureisfoss:
Now it's easy to jump to your bookmarked directories in lf, just press J and enter the shortcut.

### DIFF
--- a/.config/lf/lfrc
+++ b/.config/lf/lfrc
@@ -148,9 +148,22 @@ cmd bulkrename ${{
     lf -remote "send $id unselect"
 }}
 
+cmd shortcutjump ${{
+	set -f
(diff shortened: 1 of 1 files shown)


yushisora:
Maybe it is better to use ` --bind one:accept` with `fzf` to accept the only option left?

futureisfoss:
@yushisora I have tried it but soon realized that it can lead to accidents, your shortcuts will become muscle memory for you so you type the whole sequence even if the jump had already happened. Alternatively you could also shorten your shortcut itself if you feel like it is too long.

yushisora:
I have tested it a bit and now understand what you mean by that.

Anyway, thank you for the script. It is very useful.

PR #1455, opened 2025-03-03
Fix UI rendering issues in neovim

futureisfoss:
Neovim started showing UI rendering issues after the latest update. Some of the Unicode characters used in the vim-airline plugin were found to be the cause, this commit aims to solve the issue by removing those characters.

Check this issue for more details - https://github.com/vim-airline/vim-airline/issues/2704

### DIFF
--- a/.config/nvim/init.vim
+++ b/.config/nvim/init.vim
@@ -63,7 +63,8 @@ colorscheme vim
 	endif
 	let g:airline_symbols.colnr = ' C:'
 	let g:airline_symbols.linenr = ' L:'
-	let g:airline_symbols.maxlinenr = '☰ '
+	let g:airline_symbols.maxlinenr = ' '
+	let g:airline#extensions#whitespace#symbol = '!'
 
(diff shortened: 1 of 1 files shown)


aartoni:
Thanks for your contribution! I'm not really sure but it could be that the symbol had some specific meaning behind it. I've found two similar symbols to keep it consistent:

- [capital greek Xi](https://en.wikipedia.org/wiki/Xi_(letter)#Unicode): `Ξ`;
- [triple bar (aka equivalence)](https://en.wikipedia.org/wiki/Triple_bar): `≡`.

From the bug report it seems that [the bug has already been fixed](https://github.com/neovim/neovim/issues/31956#issuecomment-2585046161) by the upstream so you might want to revert this change as soon as Neovim 11 comes out.

futureisfoss:
@aartoni I don't really have a strong opinion about the symbols, but feel free to revert this change if upstream fixes the rendering issue.

PR #1461, opened 2025-03-25
Add support for hidden images if initial file is hidden

mackenziebasaraba:
If the initial image is hidden, the script uses ls -a to include other hidden files; otherwise, it uses the default behaviour.


mackenziebasaraba:
Accidentally sent a bashism in my first commit; zero clue why it was running on my machine without error.

PR #1463, opened 2025-03-29
maint(init.vim): use `$XDG_CONFIG_HOME`

vladdoster:
- Substitute `~/.config` with `${XDG_CONFIG_HOME:-$HOME/.config}` to be consistent with rest of `init.vim`.

### DIFF
--- a/.config/nvim/init.vim
+++ b/.config/nvim/init.vim
@@ -153,4 +153,4 @@ nnoremap <leader>h :call ToggleHiddenAll()<CR>
 " Here leader is ";".
 " So ":vs ;cfz" will expand into ":vs /home/<user>/.config/zsh/.zshrc"
 " if typed fast without the timeout.
-silent! source ~/.config/nvim/shortcuts.vim
+silent! source ${XDG_CONFIG_HOME:-$HOME/.config}/nvim/shortcuts.vim


LukeSmithxyz:
You can't use bash parameter expansion in nvim itself. I doesn't know what that means. Note the other invocations call the system shell, which should be done here.

vladdoster:
> You can't use bash parameter expansion in nvim itself. I doesn't know what that means. Note the other invocations call the system shell, which should be done here.

It seems you can as I tested in Neovim & Vim. Am I missing something?

https://github.com/user-attachments/assets/b6317c87-13e5-45d9-af1e-ac4aaa32bc2c

It is also done on [line 5](https://github.com/LukeSmithxyz/voidrice/pull/1463/files#diff-eb5fdde5c228eb3e6ded23fb3a31773ed9476ce5451e705aef9e087204f3be67R5)

```vim
silent !mkdir -p ${XDG_CONFIG_HOME:-$HOME/.config}/nvim/autoload/
```

LukeSmithxyz:
I'll check when I can. I was under the impression this wouldn't work and got an error running it. Will try later.

vladdoster:
@LukeSmithxyz, friendly ping -- did you get a chance to test this change?

PR #1469, opened 2025-07-08
Additions to rssget and make rssadd accept url tags as $2

miliukhin:
Improve rssget: now able to parse YouTube @username urls, watch urls,
and Telegram urls (feeds for Telegram are generated by rss-bridge). Make rssadd accept tags as "$2".

### DIFF
--- a/.local/bin/bookmark
+++ b/.local/bin/bookmark
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+getbookmark() {
+	bookmark="$(xclip -selection clipboard -o)"
+	tags=$(echo "" | dmenu -p "Enter bookmark tags (ENTER to save without tags)")
+}
+
+write_to_file () {
+	if grep -q -E "^$bookmark( |$)" "$file"; then
+		notify-send "Already bookmarked $bookmark!"
+	else
+		notify-send "Bookmarked $bookmark"
+		echo "$bookmark" "$tags" >> "$file"
+		echo "$bookmark" | xclip -selection clipboard -i
+	fi
+}
+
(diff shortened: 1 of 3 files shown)


.local/bin/bookmark
miliukhin, line 1: This script asks you what to bookmark via dmenu and then does the bookmarking. Allows tagging.
It bookmarks urls just like bookmarkthis, but from the clipboard (when using Vim Vixen it's very convenient to just press "y" and then the bookmarking shortcut to save the link, and there are some other cases when this way is more comfy).
For RSS feeds, it uses rssget.
One question you may have: "Why bookmark songs?" - Well, sometimes I listen to internet radio streams, and it's nice to be able to quickly save the names of the songs I like to look them up later.

Of course, dwm bindings are needed:

```
{ MODKEY|ShiftMask,		XK_b,		spawn,		{.v = (const char*[]){ "bookmark", NULL } } },
{ MODKEY,			XK_Insert,	spawn,		SHCMD("xdotool type $(grep -v '^#' ~/.local/share/larbs/snippets | dmenu -i -l 50 | cut -d' ' -f1)") },
```

I think they're identical to those in "Bookmarking for UNIX Chads" video.

ncv-xyz:
If #1499 gets merged (adds a radio script), would you consider changing the songs file from `~/.local/share/larbs/songlog` to `~/.local/share/radio-list`?

miliukhin:
> If #1499 gets merged (adds a radio script), would you consider changing the songs file from `~/.local/share/larbs/songlog` to `~/.local/share/radio-list`?

I don't think this is a good idea - your list is a list of radio urls. My list is a list of bookmarked song names. I'd keep them as separate files.

ncv-xyz:
Ah, sorry, thought it was bookmarking URLs

PR #1481, opened 2026-07-09
Add vim nav bluetooth module to statusbar

CMDR-Shrine:
This status bar module makes Bluetooth painless.

Middle click to toggle bluetooth on/off
Left click to open bluetui.

The bluetui supports **vim** bindings and looks great.

Main Deps:
- bluetui
- bluetoothctl

image

Will need a block entry.

```{"", "sb-bluetooth", 5, 20},```

### CODE REVIEW

**Comment by ncv-xyz** [comment]
```bash
timeout 0.5s bluetoothctl show | grep -q "Powered: yes" && echo "🔵BT" || echo "⚫BT"
```
Could the if block be rewritten as this?

CMDR-Shrine:
That is more elegant! echo will never fail so its safe.

PR #1484, opened 2026-07-15
Fix pauseallmpv freezing dwm

ncv-xyz:
There was a change in dwm's spawn code which leads to suspended processes freezing the desktop. See https://github.com/LukeSmithxyz/dwm/issues/252 By sending pauseallmpv's output to /dev/null dwm doesn't hang.

Closed my last PR as it doesn't fix it for Super+Shift+P

### DIFF
--- a/.local/bin/pauseallmpv
+++ b/.local/bin/pauseallmpv
@@ -5,6 +5,6 @@
 # every single one of them with one command! This is bound to super + shift + p
 # (with other things) by default and is used in some other places.
 
-for i in $(ls /tmp/mpvSockets/*); do
-	echo '{ "command": ["set_property", "pause", true] }' | socat - "$i";
+for i in /tmp/mpvSockets/*; do
+	echo '{ "command": ["set_property", "pause", true] }' | socat - "$i" >/dev/null 2>&1
 done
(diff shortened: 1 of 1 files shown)


ncv-xyz:
Forgot to add, also removed ls because it was unnecessary

Agecaltin:
Hey @ncv-xyz thanks for looking into this and fixing the issue. Just for my understanding: I noticed two days ago that my super+backlash, which calls sysact causes the dwm freeze on a lock attempt. Running sysact from the terminal worked fine. Apparently it is because of the pauseallmpv. But how come this issue now surfaced for me? I did not change the dwm source etc. Thanks for your help

Agecaltin:
Just for anyone else, this resolved the issue for me. No changes to dwm or sysact needed, just the pauseallmpv.

ncv-xyz:
@Agecaltin 

> But how come this issue now surfaced for me?

Seems when socat exits non-0 (e.g. old sockets in /tmp/mpvSockets/) dwm for
some reason hangs. Has to do with a change in dwm's code, according to the
issue I linked. I don't know from which version or how exactly. So from a 
terminal it works fine but freezes from dwm.

PR #1485, opened 2026-07-15
Remove extra printf in sb-battery

ncv-xyz:


LukeSmithxyz:
The separation of the `printf` commands is for on systems with multiple/extra batteries. You want all batteries to print on only one line, then add only one final `\\n` for cli use.

ncv-xyz:
Ah, sorry

PR #1493, opened 2026-08-22
sb-scripts replace awk with shell built-ins/parameter expansions

ncv-xyz:
inb4 calling this autism (this is)

sb-cpu time results:
```
for x in $(seq 1000); do sb-cpu >/dev/null; done  3.46s user 7.26s system 68% cpu 15.559 total # awk
for x in $(seq 1000); do sb-cpu >/dev/null; done  2.62s user 5.49s system 50% cpu 16.049 total # parameter expansion
```

sb-disk time results:
```
sh -c 'for x in $(seq 1000); do sb-disk >/dev/null; done'  2.57s user 4.24s system 122% cpu 5.560 total # awk
sh -c 'for x in $(seq 1000); do sb-disk >/dev/null; done'  0.93s user 2.21s system 96% cpu 3.257 total # parameter expansion
```

sb-kbselect time results:
```
sh -c 'for x in $(seq 1000); do sb-kbselect >/dev/null; done'  3.73s user 4.96s system 113% cpu 7.648 total # awk
sh -c 'for x in $(seq 1000); do sb-kbselect >/dev/null; done'  3.45s user 3.82s system 96% cpu 7.546 total # shell
```

sb-memory time results:
```
sh -c 'for x in $(seq 1000); do sb-memory >/dev/null; done'  3.00s user 5.72s system 133% cpu 6.530 total # awk
sh -c 'for x in $(seq 1000); do sb-memory >/dev/null; done'  1.32s user 2.69s system 95% cpu 4.180 total # parameter expansion
```

sb-tasks time results:
```
sh -c 'for x in $(seq 1000); do sb-tasks >/dev/null; done'  2.47s user 4.16s system 117% cpu 5.664 total # awk
sh -c 'for x in $(seq 1000); do sb-tasks >/dev/null; done'  1.17s user 2.15s system 109% cpu 3.023 total # shell
```

PR #1494, opened 2026-08-22
sb-news replace awk with parameter expansion

ncv-xyz:
Also cleaned up the formatting.

I cannot figure out what the purpose of running cat ~/.config/newsboat/.update was, neither newsboat nor newsup crated it while updating feeds.


ncv-xyz:
Combining this PR with #1493

PR #1495, opened 2026-08-22
sb-memory replace awk with parameter expansion

ncv-xyz:
time results:
```
sh -c 'for x in $(seq 1000); do sb-memory >/dev/null; done'  3.00s user 5.72s system 133% cpu 6.530 total # awk 
sh -c 'for x in $(seq 1000); do sb-memory >/dev/null; done'  1.32s user 2.69s system 95% cpu 4.180 total # parameter expansion
```


ncv-xyz:
Combining this PR with #1493

PR #1496, opened 2026-08-22
sb-tasks replace awk with shell built-ins

ncv-xyz:
time results:
```
sh -c 'for x in $(seq 1000); do sb-tasks >/dev/null; done'  2.47s user 4.16s system 117% cpu 5.664 total # awk 
sh -c 'for x in $(seq 1000); do sb-tasks >/dev/null; done'  1.17s user 2.15s system 109% cpu 3.023 total # shell
```


ncv-xyz:
Combining this PR with #1493

PR #1497, opened 2026-08-22
sb-disk replace awk with parameter expansion

ncv-xyz:
time results:
```
sh -c 'for x in $(seq 1000); do sb-disk >/dev/null; done'  2.57s user 4.24s system 122% cpu 5.560 total # awk
sh -c 'for x in $(seq 1000); do sb-disk >/dev/null; done'  0.93s user 2.21s system 96% cpu 3.257 total # parameter expansion
```


ncv-xyz:
Combining this PR with #1493

PR #1498, opened 2026-08-23
sb-kbselect replace awk with shell built-ins

ncv-xyz:
Also cleaned up formatting. time results (marginal): 
```
sh -c 'for x in $(seq 1000); do sb-kbselect >/dev/null; done'  3.73s user 4.96s system 113% cpu 7.648 total # awk 
sh -c 'for x in $(seq 1000); do sb-kbselect >/dev/null; done'  3.45s user 3.82s system 96% cpu 7.546 total # shell
```


ncv-xyz:
Combining this PR with #1493

PR #1501, opened 2026-08-24
transadd group commands instead of subshell

ncv-xyz:
Also replaced the if block in td-toggle with logical operators.


ncv-xyz:
Combined with #1503

PR #1503, opened 2026-08-24
General optimizations

ncv-xyz:
Commits summary:

- dmenuunicode replace with shell built-ins

Added XDG variable and replaced cut and sed.

- transadd group commands instead of subshell

Also replaced the if block in td-toggle with logical operators.

- sb-torrent replace with shell built-ins

time results:
```
sh -c 'for x in $(seq 1000); do sb-torrent >/dev/null; done'  17.98s user 17.84s system 114% cpu 31.289 total # utils
sh -c 'for x in $(seq 1000); do sb-torrent >/dev/null; done'  14.65s user 8.73s system 94% cpu 24.851 total # shell
```

- sb-popupgrade use pacman then yay

yay -Su pulls updates for AUR packages, so pacman -Syu should be used for non-AUR packages. This way the script works on systems without yay or AUR.

- linkhandler replace printf with echo

I've no clue why I piped printf in my original PR.


.local/bin/statusbar/sb-popupgrade
```diff
@@ -1,9 +1,12 @@
...
+
+sudo pacman -Syu
+
+command -v yay >/dev/null 2>&1 && yay -Su
 
-yay -Syu
 pkill -RTMIN+8 "${STATUSBAR:-dwmblocks}"
 
-printf "\\nUpgrade complete.\\nPress <Enter> to exit window.\\n\\n"
+printf "\nUpgrade complete.\nPress any key to exit window."
```
2084x, line 11: this isn't true. enter is the only single keystroke that will close the window. the prompt didn't need to change.

ncv-xyz: I thought from my testing any key caused read to exit. I'll test again and change it back in #1504

PR #1504, opened 2026-08-25
General optimizations (cont.)

ncv-xyz:
Meant to include these in my last PR, but my internet went out.

- Fix weath

Works the same as before, but prints to terminal rather than displaying a notification. Main issue was a second less here would run if the first succeeded:

```
less -S "$report" || data="$(curl -sfm 5 "${WTTRURL:-wttr.in}/$1")" && echo "$data" | less -S
```

- Simplified dmenuhandler

Same changes for linkhandler from #1482. Now exits when no feed given.

- rotdir replace awk with shell built-ins

time results:

```
sh -c 'for x in $(seq 1000); do rotdir test.png >/dev/null; done'  2.34s user 5.39s system 113% cpu 6.832 total # awk
sh -c 'for x in $(seq 1000); do rotdir test.png >/dev/null; done'  1.21s user 3.24s system 95% cpu 4.640 total # shell
```

- sb-volume reduce subshells and use printf

time results are about the same, but this spawns only one subshell instead of two.

- sb-brightness replace cat with read

time results:

```
sh -c 'for x in $(seq 1000); do sb-brightness >/dev/null; done'  1.13s user 3.38s system 90% cpu 4.957 total # cat
sh -c 'for x in $(seq 1000); do sb-brightness >/dev/null; done'  0.65s user 1.20s system 80% cpu 2.297 total # read
```

- sb-popupgrade correct exit prompt

Whoops

- Fix typos

Mistakenly put ;


ncv-xyz:
Does anyone use sb-brightness?

2084x:
since you're updating `weath` I thought I'd mention that wttr now recommends using the .is domain for scripts so might be worth making it the default now.
https://github.com/chubin/wttr.in#domains--reliability

the `?F` parameter can also be used to hide the social media follow prompt, which I can't imagine most want to see in their terminal weather reports. personally I'd add that as the default too.

